PR #1, opened 2021-03-15
fix tests

ahl:
@bnaecker can you take a look? You and I wrestled on a little bit of this code so I wanted to check in since I was doing this on  purpose ;-)


bnaecker (approved):
Thanks for catching this, and my apologies for introducing the regression. I was caught in a clippy-induced haze, I think, and accidentally ended up somewhere that worked, but did not pass the tests. Thanks!

PR #5, opened 2021-03-16
Documention improvements

### LINKED ISSUE #3: Document the probe argument types
Currently we support integral types and `char *` as a string, but this isn't documented anywhere obvious.

bnaecker:
- Explicitly lists the supported probe data types
- Includes notes about ASM feature
- Includes notes about nightly compiler

Closes #3

### DIFF
--- a/usdt/src/lib.rs
+++ b/usdt/src/lib.rs
@@ -1,9 +1,12 @@
 //! Expose USDT probe points from Rust programs.
 //!
+//! Overview
+//! --------
+//!
 //! This crate provides methods for compiling definitions of DTrace probes into Rust code, allowing
 //! rich, low-overhead instrumentation of Rust programs.
 //!
-//! Users define a provider, with one or more probe functions, in the D language. For example:
+//! Users define a _provider_, with one or more _probe_ functions, in the D language. For example:
 //!
 //! ```d
 //! provider test {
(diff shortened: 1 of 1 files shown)


usdt/src/lib.rs
```diff
@@ -12,20 +15,50 @@
...
+//! names are intended to help understand the behavior of a program, so they should be semantically
+//! meaningful. Probes accept zero or more arguments, data that is associated with the probe event
+//! itself (timestamps, file descriptors, filesystem paths, etc.). The arguments may be specified
+//! as any of the exact bit-width integer types (e.g., `int16_t`) or strings (`char *`s). See
+//! [Data types](#data-types) for a full list of supported types.
+//!
 //! Assuming the above is in a file called `"test.d"`, this may be compiled into Rust code with:
 //!
 //! ```ignore
+//! #![feature(asm)]
```
ahl, line 28: where does this need to appear? mod.rs? lib.rs? main.rs?

usdt/src/lib.rs
```diff
@@ -12,20 +15,50 @@
...
 //! usdt::dtrace_provider!("test.d");
 //! ```
 //!
-//! This procedural macro will return a Rust macro for each probe defined in the provider. For
-//! example, one may then call the `start` probe via:
+//! This procedural macro will return a Rust macro for each probe defined in the provider. Note
```
ahl, line 32:
```suggestion
//! This procedural macro will generate a Rust macro for each probe defined in the provider. Note
```

usdt/src/lib.rs
```diff
@@ -34,19 +67,66 @@
...
+//! - `char *`
+//!
+//! Currently, up to six (6) arguments are supported, though this limitation may be lifted in the
+//! future.
+//!
+//! Registration
+//! ------------
+//!
+//! USDT probes must be registered with the DTrace kernel module. This is done in Rust code with
+//! the [`register_probes`] function, which must be called before any of the probes become
```
ahl, line 94:
```suggestion
//! USDT probes must be registered with the DTrace kernel module. This is done via a call to
//! the [`register_probes`] function, which must be called to register those probes with DTrace
```

ahl (approved):
looks good. some minor suggested tweaks

usdt/src/lib.rs
bnaecker: Good question, I'm not sure yet. I'm running some experiments to find out.

bnaecker: After some experimentation, here's what I've found. Looks like the `!#[feature(asm)]` directive must be at the crate root. The story around the proc-macro invocation site is more complex, since it generates macros. There's at least one improvement that we can do (getting the full path of the probe and is-enabled probe in the proc-macro), but practically speaking, `usdt::dtrace_provider!` should also be invoked at the crate root.

PR #10, opened 2021-03-18
simplify use of probes with zero arguments

ahl:
This allows a probe with zero arguments to be invoked without a closure (optionally).

### DIFF
--- a/usdt-impl/src/no-linker.rs
+++ b/usdt-impl/src/no-linker.rs
@@ -90,10 +90,18 @@ fn compile_probe(
         })
         .collect::<Vec<_>>();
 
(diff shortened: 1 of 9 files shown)


usdt-impl/src/no-linker.rs
```diff
@@ -104,6 +112,7 @@ fn compile_probe(
     let out = quote! {
         #[allow(unused)]
         macro_rules! #macro_name {
+            #no_args
```
bnaecker, line 115: I might rename this to "#no_args_branch` or `#no_args_branch` or something, to indicate it's really a (possibly-empty) separate matcher/transcriber arm of the macro.

PR #12, opened 2021-03-18
Adds no-linker feature for building on macos without relying on the

bnaecker:
platform dynamic linker

- Adds the `"no-linker"` feature flag to the `usdt` and `usdt-impl`
crates, forwarding the former to the latter
- Better conditional compilation of the `usdt_impl::internal` module,
based on the platform and new set of feature flags
- Combines some of the shared code used to generate the USDT macros
themselves into a `usdt-impl/src/common.rs` mod. The actual
implementation of the macro in the macos/linker case has been updated.
Previously, we were directly calling the (linker-defined) probe function
by creating a Rust fn with link-name of that mangled probe function.
This mapped directly onto the "usual" C implementation. That has been
updated to marshal the arguments just as we do on no-linker platforms,
and use the x86 call instruction directly. This all works, but it is a
difference, with the goal of being able to share code and test
everything more directly.
- Implements the linker/no-linker versions using the new shared code
- Adds tests for the new shared code
- Adds support for correctly extracting the module name in the no-linker
version of the code. This works by returning the containing file name in
addition to the function name from the `addr_to_info` function, and then
passing that as the module name in the DOF sent via ioctl(2).
- Adds simple test for the `asm_rec` function

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,6 +36,7 @@ jobs:
     strategy:
       matrix:
         os: [ macos-latest ]
+        features: [ "", "--features no-linker" ]
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
@@ -43,4 +44,4 @@ jobs:
           toolchain: nightly
           override: true
           profile: minimal
-      - run: RUST_BACKTRACE=1 cargo test --release --no-fail-fast --verbose
+      - run: RUST_BACKTRACE=1 cargo test --release --no-fail-fast ${{ matrix.features }} --verbose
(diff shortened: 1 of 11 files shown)


ahl (commented):
What if we put even more into common so that the common code looked like this:

```rust
    let out = quote! {
        #[allow(unused)]
        macro_rules! #macro_name {
            #no_args_match
            ($args_lambda:expr) => {
                // NOTE: This block defines an internal empty function and then a lambda which
                // calls it. This is all strictly for type-checking, and is optimized out. It is
                // defined in a scope to avoid multiple-definition errors in the scope of the macro
                // expansion site.
                {
                    fn _type_check(#(#type_check_args),*) { }
                    let _ = || {
                        #preamble
                        _type_check(#(#expanded_lambda_args),*);
                    };
                }

         // DO THE ASM OR WHATEVER GETS PASSED IN
...
```

Even `empty.rs` could basically call that with the inner mechanism of `quote!{}`

Issue #4, opened 2021-03-16
test and validate compile-time mistakes


ahl:
We should have a collection of tests that represent developer errors and ensure that the error messages generated are helpful and reflective of our intentions. For example:

- probe with unsupported parameter type (tell the developer what is supported)
- incorrect parameter type for probe input (sensible type error)
- raw parameters or tuple passed to probe rather than function / closure

bnaecker:
The first of these is partially improved by #7

bnaecker:
Closed by #13

PR #13, opened 2021-03-18
Adds tests to verify useful error messages when compiling fails

bnaecker:
- Adds a suite of tests which verify the compiler error messages when
the `usdt` crate is misused in various likely ways. This includes a
provider definition with an unsupported type, a type mismatch in the
probe macro invocation, calling the probe macro without a closure, and
specifying a non-existent provider file to the proc macro or build.rs.
- Adds back the `asm` feature as a default and updates the CI config
accordingly.
- More appropriate conditional compilation of the actual internal
implementation file under various combinations of platform a feature
flags.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,7 +23,6 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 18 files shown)


bnaecker:
This will close #4 when merged

PR #14, opened 2021-03-18
more complex cmd/library example + making it work on illumos

ahl:


usdt-impl/src/no-linker.rs
```diff
@@ -100,12 +110,13 @@ fn extract_probe_records_from_section() -> Result<Option<Section>, crate::Error>
...
-    // Without this the illumos linker may decide to omit symbols referencing this section.
-    // The macos linker doesn't seem to require this.
+    // Without this the illumos linker may decide to omit the symbols above that
+    // denote the start and stop addresses for this section. The macos linker
+    // doesn't seem to require this.
     #[cfg(target_os = "illumos")]
     #[link_section = "set_dtrace_probes"]
     #[used]
-    static FORCE_LOAD: [u8; 0] = [];
+    static FORCE_LOAD: [u64; 0] = [];
```
bnaecker, line 119: Is the change to `u64` required? I'm curious if the linker behaves differently in either case.

ahl: This is not necessary. I made this change because it effects sh_addralign and I thought it would be better if both the records and the placeholders had the same value there.

bnaecker (approved):

ahl:
It turns out this doesn't work when you have multiple probe call sites because `FORCE_LOAD` is multiply defined.

bnaecker:
> It turns out this doesn't work when you have multiple probe call sites because `FORCE_LOAD` is multiply defined.

Would it work to put this in the `#pre_macro_block` portion of the generated code? That seems to prevent multiply-defined symbols on macOS, when referring to the extern functions that the linker resolves.

ahl:
That's a really good idea. It doesn't seem to work, unfortunately, for illumos builds with --release -- possibly due to a linker bug that we're looking into.

Issue #2, opened 2021-03-16
Minimize the number of dependencies for consumers


ahl:
The dependency chain for usdt is pretty long. Some of this probably can't be avoided, but we can move the commands in `dof` and `usdt` into their own crates and/or examples to remove some of them.

ahl:
Closed by #15

PR #15, opened 2021-03-19
Prunes the dependency tree

bnaecker:
- Removes extraneous structopt dep from `dof` crate
- Adds a "des" feature to many of the packages. This is used to gate
inclusion of the tools for deserializing the probe records or DOF itself
from an object file. The effect on the dependency tree is to make the
large `goblin` dependency optional.
- Adds a small, separate binary crate `dusty`, which has the binary for
actually reading the DOF or ASM records from an object file. This is
really just the former `usdt/src/main.rs`, with appropriate changes to
the feature flags and depdendency tree.
- Note that the `"dof/des"` flag is not a default feature. It is opted
into via the `"usdt/des"` feature, which is _not_ implied by the "asm"
and "no-linker" features of the same crate. This allows the
deserialization tooling to be opted _out_ of, e.g., by most users of the
`usdt` crate. Note that the usdt::record module is conditionally
included on the basis of platform and the no-linker or asm features.

### DIFF
--- a/usdt-impl/src/record.rs
+++ b/usdt-impl/src/record.rs
@@ -5,15 +5,18 @@
 use std::{
     collections::BTreeMap,
     ffi::CStr,
-    fs,
-    path::Path,
     ptr::{null, null_mut},
 };
 
+#[cfg(feature = "des")]
+use std::{fs, path::Path};
+
 use byteorder::{NativeEndian, ReadBytesExt};
 use dof::{Probe, Provider, Section};
-use goblin::elf;
 use libc::{c_void, Dl_info};
+
+#[cfg(feature = "des")]
+use goblin::Object;
 
 pub(crate) const PROBE_REC_VERSION: u8 = 1;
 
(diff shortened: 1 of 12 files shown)


bnaecker:
This should address #2 when merged, at least partially. The number of dependencies required when building the main `usdt` crate, using the default features is reduced from over 100 to about 40, depending slightly on the platform.

usdt-impl/src/record.rs
```diff
@@ -5,90 +5,125 @@
...
-                    Ok(name) => {
-                        if name == "set_dtrace_probes" {
-                            Some(Ok(header))
-                        } else {
-                            None
+    match Object::parse(&data).map_err(|_| crate::Error::InvalidFile)? {
+        Object::Elf(object) => {
+            // Try to find our special `set_dtrace_probes` section from the section headers. These may not
+            // exist, e.g., if the file has been stripped. In that case, we look for the special __start
+            // and __stop symbols themselves.
```
ahl, line 34: I haven't seen this specific pathology -- have you see an instance where the probe section is gone entirely? The errors I've seen produce a zero-sized section.

ahl (approved):
This is great. What's `cargo tree` look like now?

usdt-impl/src/record.rs
bnaecker: No, this was unrelated to the empty-section issue. I was just being defensive against the possibility of a stripped binary.

bnaecker:
I wanted to summarize the effects of this on consuming crates. There are two cases to consider: (1) a simple binary, and (2) a library that exposes probe points to a consuming binary. The `cargo tree` output is a bit unhelpful here. Instead, I think it's more useful to look at what happens when one builds a crate. How many dependencies does that require?

On the current `master` (c4ea31b4), building the `probe-test-macro` example compiles 67 total crates. On this PR branch (b17a9ff7), it requires 36.

Let's examine the situation for our new test case `tests/fake-cmd`. This uses `tests/fake-lib` as a dependency, which is where the actual probes are defined. `fake-cmd` is a main function that just registers `fake-lib`'s probes, calls the probe via the `fake_lib::dummy` function, and then exits. In `master` / c4ea31b4, this requires 70 crates, while the PR branch pulls in 39.

PR #17, opened 2021-03-19
Adds integration test verifying that a probe is visible via `dtrace(1)`

bnaecker:
- Adds the `tests/does-it-work` crate to the workspace. This builds a
single probe, and then runs DTrace to verify that that probe is listed
with the correct properties. This relies on passwordless sudo, available
on GitHub's CI, or that the test itself is actually run with `sudo
cargo`.
- Removes stray println in the extraction of probe information from an
object file, left over from development.
- Bolds the comment in the README that the generated macros should be
included in the crate root.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
 	"probe-test-build",
 	"probe-test-macro",
 	"tests/compile-errors",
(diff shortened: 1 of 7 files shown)


ahl:
this is cool; thanks for doing this.

PR #19, opened 2021-03-26
update the path where we look for the provider.d file to appease rust-analyzer

ahl:
@pfmooney observed that rust-analyzer runs builds from a different cwd than if one runs `cargo build` by hand. Consequently, the `dtrace_provider!` macro looks for the provider file in the wrong place. This PR uses `CARGO_MANIFEST_DIR` as the base directory in which we look for the specified provider file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ which is nearly identical to the above example. However, there is no build.rs sc
 so in place of the `include!` macro, one finds the procedural macro:
 
 ```rust
-dtrace_provider!("probe-test-macro/test.d");
+dtrace_provider!("test.d");
 ```
 
 This macro generates the same macros as seen above, but does at the time the source
--- a/usdt-macro/src/lib.rs
+++ b/usdt-macro/src/lib.rs
@@ -33,9 +33,9 @@ use usdt_impl::compile_providers;
 /// in this case would be in the same directory as `"Cargo.toml"`.
 ///
 /// By default probe macros are named `{provider}_{probe}!`. Arguments are passed
-/// via a closure (or function) that returns a tuple. Note that the provided
-/// closure is only evaluated when the probe is enabled. One can then add points of
-/// instrumentation by invoking the macro:
+/// via a closure that returns a tuple. Note that the provided closure is only
+/// evaluated when the probe is enabled. One can then add points of instrumentation
+/// by invoking the macro:
 ///
 /// ```ignore
 /// fn do_stuff(count: u8, name: String) {
--- a/usdt-macro/src/lib.rs
+++ b/usdt-macro/src/lib.rs
@@ -1,8 +1,8 @@
 //! Prototype proc-macro crate for parsing a DTrace provider definition into Rust code.
 // Copyright 2021 Oxide Computer Company
 
-use std::fs;
 use std::iter::FromIterator;
+use std::{fs, path::Path};
 
 use quote::quote;
 use syn::{parse_macro_input, Lit};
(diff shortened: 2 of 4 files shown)


usdt-macro/src/lib.rs
```diff
@@ -79,10 +79,11 @@ pub fn dtrace_provider(item: proc_macro::TokenStream) -> proc_macro::TokenStream
         _ => panic!("DTrace provider must be a single literal string filename"),
     };
     let source = if filename.ends_with(".d") {
-        fs::read_to_string(&filename).expect(&format!(
+        let dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+        let path = Path::new(&dir).join(&filename);
```
pfmooney, line 83: If that env variable isn't available, would it make sense to attempt to use the bare filename as the path?  I'm not sure if anyone would attempt to use this by building with `rustc(1)` directly.

bnaecker: Doesn't hurt to make this `unwrap_or_else` or similar.

README.md
```diff
@@ -136,7 +136,7 @@ which is nearly identical to the above example. However, there is no build.rs sc
 so in place of the `include!` macro, one finds the procedural macro:
 
 ```rust
-dtrace_provider!("probe-test-macro/test.d");
```
bnaecker, line 139: It seems like a good idea to put a note in the README about how this file is found. Just a quick parenthetical like "(The file must be at a path relative to the `CARGO_MANIFEST_DIR`)".

bnaecker (approved):
Small suggestion about adding a note in the README, but otherwise LGTM. Thanks for handling this!

usdt-macro/src/lib.rs
```diff
@@ -31,7 +29,13 @@ use usdt_impl::compile_providers;
 /// dtrace_provider!("test.d");
 /// ```
 ///
-/// One can then instrument the application or library as one might expect:
+/// The macro looks for the file relative to the root of the package, so `"test.d"`
+/// in this case would be in the same directory as `"Cargo.toml"`.
+///
+/// By default probe macros are named `{provider}_{probe}!`. Arguments are passed
+/// via a closure (or function) that returns a tuple. Note that the provided
```
bnaecker, line 36: Actual functions aren't supported right now. This is one of the sharp-edges of the macro: there's no matcher that identifies both closures and functions, but that also filters out other paths, like `std::collections::Vec`. So it does _have_ to be a closure right now.

bnaecker (approved):
Looks great, one nit about closures vs functions. But otherwise LGTM!

README.md
pfmooney: > The file must be at a path relative to the `CARGO_MANIFEST_DIR`

Perhaps reference that it's the crate root (and the env)?  I'm not sure `CARGO_MANIFEST_DIR` is meaningful to everyone.  (I certainly didn't know about it until it was mentioned)

ahl: Agreed; I added docs that didn't explicitly reference `CARGO_MANIFEST_DIR` because I didn't know what it was until I started dumping out `std::env::vars()`

usdt-macro/src/lib.rs
ahl: What wouldn't `expr` work for functions? I tried removing the `tt` pattern and using a function just worked. What cases are we trying to catch here?

bnaecker: I could be wrong, but the goal was to allow closures, but disallow things like the actual arguments. Expr does work for functions, but it's overly permissive, as it allows a tuple with the arguments. So I put a token-tree matcher first, to filter out the tuple, which also matches a function.

So one option would be to match on an expression or a path, which matches things like `std::collections::Vec` as well as functions. The path and expr matches would expand to the same branch, and the token-tree would filter out things like tuples. I don't know that any of this needs to happen now, but if/when we want to support functions, this might be one approach.

[Here's a link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=497c610ed59038dd2eeea56c0a6e8b07) to the playground showing this in action, for future reference.

ahl: Agreed; we can defer doing anything more permissive here.

ahl:
@bnaecker no rush, but can you publish a new version on crates.io when you have a moment?

README.md
bnaecker: Nice resolution, thanks

PR #20, opened 2021-05-24
Avoid issuing ioctl on bad fd

smklein:
Before:

If `/dev/dtrace/helper` was missing, due to (for example) a missing package, probe registration would return a "bad fd" error, after attempting to issue an ioctl to a defunct file descriptor.

After:

Well, if `/dev/dtrace/helper` is gone, things are still gonna fail, but the error will be slightly more comprehensible - it should fail due to "missing file/device", which is more accurate.

### DIFF
--- a/tests/compile-errors/src/no-closure.stderr
+++ b/tests/compile-errors/src/no-closure.stderr
@@ -7,4 +7,4 @@ error: USDT probe macros should be invoked with a closure returning the argument
 7 |     mismatch_bad!(arg);
   |     ------------------- in this macro invocation
   |
-  = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
+  = note: this error originates in the macro `mismatch_bad` (in Nightly builds, run with -Z macro-backtrace for more info)
--- a/tests/compile-errors/src/type-mismatch.stderr
+++ b/tests/compile-errors/src/type-mismatch.stderr
@@ -7,4 +7,4 @@ error[E0308]: mismatched types
 7 |     mismatch_bad!(|| (bad));
   |     ------------------------ in this macro invocation
   |
-  = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
+  = note: this error originates in the macro `mismatch_bad` (in Nightly builds, run with -Z macro-backtrace for more info)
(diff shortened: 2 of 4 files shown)


bnaecker (approved):
Sweet, nice catch!

bnaecker:
AFAICT, those failing tests look like a small difference in the compiler output during errors. Took me a while to spot it, but the name of macro is now printed in error messages. E.g., 

```
note: this error originates in a macro
```

is now:

```
note: this error originates in the macro `mismatch_bad`
```

@smklein Can you overwrite expected test output with `TRYBUILD=overwrite cargo test` locally, and push those new changes?

smklein:
> AFAICT, those failing tests look like a small difference in the compiler output during errors. Took me a while to spot it, but the name of macro is now printed in error messages. E.g.,
> 
> ```
> note: this error originates in a macro
> ```
> 
> is now:
> 
> ```
> note: this error originates in the macro `mismatch_bad`
> ```
> 
> @smklein Can you overwrite expected test output with `TRYBUILD=overwrite cargo test` locally, and push those new changes?

Errr this 

> AFAICT, those failing tests look like a small difference in the compiler output during errors. Took me a while to spot it, but the name of macro is now printed in error messages. E.g.,
> 
> ```
> note: this error originates in a macro
> ```
> 
> is now:
> 
> ```
> note: this error originates in the macro `mismatch_bad`
> ```
> 
> @smklein Can you overwrite expected test output with `TRYBUILD=overwrite cargo test` locally, and push those new changes?

I did this (on both Linux / illumos) - I'm not seeing any test differences. Is this a mac-specific thing?

bnaecker:
Yeah, looks like the tests have now passed on Linux. I'll look into it on my mac, definitely odd.

ahl:
Nice catch; thanks, @smklein !

bnaecker:
Wacky. It passes fine on my local machine, so something about CI. I'll see what I can find.

bnaecker:
So @smklein I don't see any changes related to the test files. Did you run `cargo test --workspace`? or just `cargo test`? The former is where those tests come from. I can repro this when running that way on my Mac. I'll try now on my Linux machine.

bnaecker:
And apologies for not making clear that that's required, if that's where the confusion came from!

bnaecker:
Sweet, I've repro'd this on Linux and Mac, seems the `--workspace` flag wasn't passed when testing. I'll add a note in the README and include the new compiler error messages. Thanks @smklein!

PR #21, opened 2021-06-08
Adds ARM registers for building probe arguments.

bnaecker:
This should support all ARM architectures, including Apple's M1.

### DIFF
--- a/usdt-impl/src/common.rs
+++ b/usdt-impl/src/common.rs
@@ -41,8 +41,16 @@ pub fn generate_type_check(types: &[dtrace_parser::DataType]) -> TokenStream {
 // Return code to destructure a probe arguments into identifiers, and to pass those to ASM
 // registers.
 pub fn construct_probe_args(types: &[dtrace_parser::DataType]) -> (TokenStream, TokenStream) {
-    // TODO this will fail with more than 6 parameters.
+    // x86_64 passes the first 6 arguments in registers, with the rest on the stack.
+    // We limit this to 6 arguments in all cases for now, as handling those stack
+    // arguments would be challenging with the current `asm!` macro implementation.
+    #[cfg(target_arch = "x86_64")]
     let abi_regs = ["rdi", "rsi", "rdx", "rcx", "r8", "r9"];
+    #[cfg(target_arch = "aarch64")]
+    let abi_regs = ["x0", "x1", "x2", "x3", "x4", "x5"];
+    #[cfg(not(any(target_arch = "aarch64", target_arch = "x86_64")))]
+    compile_error!("USDT only supports x86_64 and ARM64 architectures");
+
     assert!(
         types.len() <= abi_regs.len(),
         "Up to 6 probe arguments are currently supported"
--- a/usdt-impl/src/linker.rs
+++ b/usdt-impl/src/linker.rs
@@ -152,13 +152,20 @@ fn compile_probe(
         }
     };
 
+    #[cfg(target_arch = "x86_64")]
+    let call_instruction = quote! { "call {probe_fn}" };
+    #[cfg(target_arch = "aarch64")]
+    let call_instruction = quote! { "bl {probe_fn}" };
+    #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+    compile_error!("USDT only supports x86_64 and AArch64 architectures");
+
     let impl_block = quote! {
         unsafe {
             if $crate::#mod_name::#is_enabled_fn() != 0 {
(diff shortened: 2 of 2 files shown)


usdt-impl/src/common.rs
```diff
@@ -42,7 +42,11 @@ pub fn generate_type_check(types: &[dtrace_parser::DataType]) -> TokenStream {
 // registers.
 pub fn construct_probe_args(types: &[dtrace_parser::DataType]) -> (TokenStream, TokenStream) {
     // TODO this will fail with more than 6 parameters.
-    let abi_regs = ["rdi", "rsi", "rdx", "rcx", "r8", "r9"];
+    let abi_regs = if cfg!(any(target_arch = "arm64", target_arch = "aarch64")) {
+        ["r0", "r1", "r2", "r3", "r4", "r5"]
+    } else {
+        ["rdi", "rsi", "rdx", "rcx", "r8", "r9"]
```
ahl, line 48: now that I think about it, it might warrant a comment here:
```rust
// x86-64 passes the first 6 function parameters in registers; subsequent parameters are passed on the stack. We currently limit ourselves to 6 probe parameters as navigating stack-passed parameters with the current semantics of the asm! macro would be challenging
```

Also what do you think about:

```rust
let abi_regs = if cfg!(target_arch = "x86_64") {
...
} else if cfg!(any(target_arch = "arm64", target_arch = "aarch64")) {
...
} else {
// Suddenly the Dungeon collapses!! - You die...
// i.e. a compiler error
}
```

bnaecker: Good call, thank you!

usdt-impl/src/common.rs
```diff
@@ -42,7 +42,14 @@ pub fn generate_type_check(types: &[dtrace_parser::DataType]) -> TokenStream {
...
 pub fn construct_probe_args(types: &[dtrace_parser::DataType]) -> (TokenStream, TokenStream) {
     // TODO this will fail with more than 6 parameters.
-    let abi_regs = ["rdi", "rsi", "rdx", "rcx", "r8", "r9"];
+    let abi_regs = if cfg!(target_arch = "x86_64") {
+        // x86 is first as that provides the stricter limit on the number of arguments
+        ["rdi", "rsi", "rdx", "rcx", "r8", "r9"]
+    } else if cfg!(any(target_arch = "arm64", target_arch = "aarch64")) {
+        ["x0", "x1", "x2", "x3", "x4", "x5"]
+    } else {
+        compile_error!("USDT only supports x86_64 and ARM64 architectures");
```
luqmana, line 51: This line is always going to be tripped at compile time. You need to use `#[cfg(..)]` attributes instead as those will actually work as expected with the conditional compilation here.

usdt-impl/src/common.rs
```diff
@@ -42,7 +42,14 @@ pub fn generate_type_check(types: &[dtrace_parser::DataType]) -> TokenStream {
 // registers.
 pub fn construct_probe_args(types: &[dtrace_parser::DataType]) -> (TokenStream, TokenStream) {
     // TODO this will fail with more than 6 parameters.
-    let abi_regs = ["rdi", "rsi", "rdx", "rcx", "r8", "r9"];
+    let abi_regs = if cfg!(target_arch = "x86_64") {
+        // x86 is first as that provides the stricter limit on the number of arguments
+        ["rdi", "rsi", "rdx", "rcx", "r8", "r9"]
+    } else if cfg!(any(target_arch = "arm64", target_arch = "aarch64")) {
```
luqmana, line 48: AFAIK Rust only uses `aarch64` though I suppose checking both can't hurt.

usdt-impl/src/common.rs
luqmana: Something like:
```Rust
    #[cfg(target_arch = "x86_64")]
    const ABI_REGS: &[&str] = &["rdi", "rsi", "rdx", "rcx", "r8", "r9"];

    #[cfg(target_arch = "aarch64")]
    const ABI_REGS: &[&str] = &["x0", "x1", "x2", "x3", "x4", "x5"];

    #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
    compile_error!("USDT only supports x86_64 and ARM64 architectures");
```

bnaecker: Thanks @luqmana I see this, now:

> cfg!, unlike #[cfg], does not remove any code and only evaluates to true or false. For example, all blocks in an if/else expression need to be valid when cfg! is used for the condition, regardless of what cfg! is evaluating.

PR #23, opened 2021-10-05
fix issues relating to data not living long enough

ahl:
This expansion of one of the macros may best exemplify this change (hand re-formatted):

```rust
macro_rules ! test_stop {
    ($tree:tt) => {
        compile_error!("USDT probe macros should be invoked with a closure returning the arguments");
    };
    ($args_lambda:expr) => {
        {
            fn _type_check<S: AsRef<str>>(_: S, _: u8, _ :S) { }
            let _ = || {
                let args = $args_lambda();
                _type_check(args.0, args.1, args.2);
            };
        }
        unsafe {
            if $crate::__usdt_private_test::test_stop_enabled() != 0 {
                let args = $args_lambda();
                let arg_0 = [(args.0.as_ref() as &str).as_bytes(), &[0_u8]].concat();
                let arg_1 = (args.1 as i64);
                let arg_2 = [(args.2.as_ref() as &str).as_bytes(), &[0_u8]].concat();

                asm!(
                    ".reference {typedefs}",
                    "call {probe_fn}",
                    ".reference {stability}",
                    typedefs = sym $crate::__usdt_private_test::typedefs,
                    probe_fn = sym $crate::__usdt_private_test::test_stop,
                    stability = sym $crate::__usdt_private_test::stability,
                    in ("rdi") (arg_0.as_ptr() as i64),
                    in ("rsi") (arg_1),
                    in ("rdx") (arg_2.as_ptr() as i64),
                    options (nomem, nostack, preserves_flags),
                );
            }
        }
    };
}
```

Note in particular: we now store the converted value rather than just the `i64` value of the pointer. Storing only the numerical value didn't force the actual value to live long enough so the compiler could easily overwrite it.

Further, we now allow any `AsRef<str>` for a string value which we then convert (`as_ref().as_bytes()` etc)

To the careful reader: yes, this means that this mostly wasn't working / was working by accident. In particular, it would readily fail with dynamic strings .

### DIFF
--- a/usdt-impl/src/common.rs
+++ b/usdt-impl/src/common.rs
@@ -7,11 +7,18 @@ use quote::{format_ident, quote};
 // Construct function call that is used internally in the UDST-generated macros, to allow
 // compile-time type checking of the lambda arguments.
 pub fn generate_type_check(types: &[dtrace_parser::DataType]) -> TokenStream {
+    let mut has_strings = false;
     let type_check_args = types
         .iter()
-        .map(|typ| {
-            let arg = syn::parse_str::<syn::FnArg>(&format!("_: {}", typ.to_rust_type())).unwrap();
-            quote! { #arg }
+        .map(|typ| match typ {
+            dtrace_parser::DataType::String => {
+                has_strings = true;
+                quote! { _: S }
+            }
+            _ => {
+                let arg = format_ident!("{}", typ.to_rust_type());
+                quote! { _: #arg }
+            }
         })
         .collect::<Vec<_>>();
     let expanded_lambda_args = (0..types.len())
@@ -23,13 +30,19 @@ pub fn generate_type_check(types: &[dtrace_parser::DataType]) -> TokenStream {
 
     let preamble = unpack_argument_lambda(&types);
 
+    let string_generic = if has_strings {
+        quote! { <S: AsRef<str>> }
+    } else {
+        quote! {}
+    };
+
     // NOTE: This block defines an internal empty function and then a lambda which
     // calls it. This is all strictly for type-checking, and is optimized out. It is
     // defined in a scope to avoid multiple-definition errors in the scope of the macro
     // expansion site.
     quote! {
         {
-            fn _type_check(#(#type_check_args),*) { }
+            fn _type_check #string_generic (#(#type_check_args),*) { }
             let _ = || {
                 #preamble
                 _type_check(#(#expanded_lambda_args),*);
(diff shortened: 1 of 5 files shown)


usdt-impl/src/common.rs
```diff
@@ -7,11 +7,18 @@ use quote::{format_ident, quote};
 // Construct function call that is used internally in the UDST-generated macros, to allow
 // compile-time type checking of the lambda arguments.
 pub fn generate_type_check(types: &[dtrace_parser::DataType]) -> TokenStream {
+    let mut has_strings = false;
```
bnaecker, line 10: I might return a tuple of `(TokenStream, bool)` from the match statement, rather than mutating this, but not crucial.

usdt-impl/src/common.rs
```diff
@@ -91,13 +110,21 @@ fn unpack_argument_lambda(types: &[dtrace_parser::DataType]) -> TokenStream {
     }
 }
 
-// Convert a supported data type to one passed to the probe function in a register
-fn asm_type_convert(typ: &dtrace_parser::DataType, input: TokenStream) -> TokenStream {
+// Convert a supported data type to a type to store for the duration of the
+// probe invocation and 2. a transformation for compatibility with an asm
+// register.
```
bnaecker, line 115: A `1.` before the "type to store..." would be helpful, so the `2.` doesn't come out of left field :)

usdt-impl/src/common.rs
```diff
@@ -140,23 +167,17 @@ mod tests {
...
         let types = &[dtrace_parser::DataType::U8, dtrace_parser::DataType::String];
+        let expected = quote! {
+            {
+                fn _type_check<S: AsRef<str>>(_: u8, _: S) { }
+                let _ = || {
+                    let args = $args_lambda();
+                    _type_check(args.0, args.1);
+                };
+            }
+        };
```
bnaecker, line 178: Nice, much better than string-munging :)

bnaecker (approved):
💯 This is awesome. Thanks for helping catch this, and for fixing it so quickly! Couple of nits, but otherwise LGTM.

usdt-impl/src/common.rs
bnaecker: Nvm, I see now that this is inside a map. We'd have to reduce this outside, which might not be any easier or clearer. Disregard :)

ahl: So that would l look like: `type_check_args.iter().map(..).unzip()` and then `out.1.iter().reduce(|a, b|a || b)` I think? ... and that seemed a little less clear. What do you think?

bnaecker: Yeah, definitely no better, and probably a good bit worse :)

usdt-impl/src/common.rs
ahl: yeesh.. not sure how I missed that

usdt-impl/src/common.rs
ahl: ok. glad you like it. I wasn't sure if I was undoing some subtle intention...

PR #25, opened 2021-10-06
buildomat: add basic build/test job mimicking the Actions workflow

jclulow:
I've put together a buildomat job that should be essentially the same as the build-and-test on Mac OS X.  I had to amend the test suite to allow me to request the use of `pfexec` in place of `sudo`.

The test is currently failing: https://buildomat.eng.oxide.computer/0/details/01FH9K91GX8BEHQH356DEWSKN7/0nrithm6crl5as6fakgzshsqve8pe8x7jheaeek7c47hi1vh/01FH9K98JMP6S7KJ5F5ME1EBGT#S411


.github/buildomat/jobs/build-and-test.sh
```diff
@@ -0,0 +1,21 @@
...
+set -o xtrace
+
+cargo --version
+rustc --version
+
+export RUST_BACKTRACE=1
+
+banner test
+ptime -m cargo test --release --no-fail-fast --verbose --workspace \
+    --features no-linker
```
bnaecker, line 21: Apologies for the confusion about this feature. It's not really documented, and intentionally so. It was intended to provide developers a knob to compile and test cross-platform bits, rather than as a true public feature flag. It's definitely not required to build the fully-functional crate on illumos systems. They're kinda difficult to follow, but [this set of conditional-compilation directives](https://github.com/oxidecomputer/usdt/blob/014f85ac97d0d923c17e5dcd51b9293e51acef7a/usdt-impl/src/lib.rs#L16-L25) results in using the `no-linker.rs` implementation file on illumos, even without specifying the actual flag.

tests/does-it-work/src/main.rs
```diff
@@ -81,15 +81,19 @@ mod tests {
...
-        let function = parts.next().expect("Expected a function name");
-        assert!(
-            function.contains("[does_it_work::run_test::"),
-            "Function name appears incorrect: {}",
-            function
-        );
-
-        let probe = parts.next().expect("Expected a probe name");
-        assert_eq!(probe, "work", "Probe name appears incorrect: {}", probe);
+        // These parts are currently specific to Mac OS X:
```
bnaecker, line 84: I'd really just prefer remove this block, actually. I've realized it's testing a platform-specific flavor of the output, which seems to demangle the function name automatically. The test adds nothing beyond the surrounding bits, and I think the conditional-compilation will raise more questions than it's worth.

tests/does-it-work/src/main.rs
```diff
@@ -23,16 +23,16 @@ fn run_test(rx: std::sync::mpsc::Receiver<()>) {
...
     use std::thread;
+    use usdt_tests_common::root_command;
 
     #[test]
     fn test_does_it_work() {
         let (send, recv) = channel();
         let thr = thread::spawn(move || run_test(recv));
-        let dtrace = Command::new("sudo")
-            .arg("dtrace")
+        let dtrace = root_command("dtrace")
```
bnaecker, line 35: I think as a result of sneaking my PR implementing JSON support in before this one, there's a similar catch in `tests/test-json/src/main.rs:99` ([here](https://github.com/oxidecomputer/usdt/blob/014f85ac97d0d923c17e5dcd51b9293e51acef7a/tests/test-json/src/main.rs#L99)). Would you mind changing that command invocation to `root_command` as well?

bnaecker (approved):
This looks great! Thanks for incorporating illumos support, definitely invaluable to have a non-macOS platform with native DTrace support. I have a few small comments, but otherwise thanks a bunch for sending this!

tests/does-it-work/src/main.rs
jclulow: I think I already got that one?

https://github.com/oxidecomputer/usdt/pull/25/files#diff-a1d523124420282a39e66d1a4de2a848cf274c4958c3efe0a1d657dbbfe2b687R100

tests/does-it-work/src/main.rs
jclulow: Sold!

PR #27, opened 2021-10-07
Fixes bad dependencies related to the dof crate

bnaecker:
We previously tried to make the `dof` crate optional, but this wasn't
quite right. It's optional if the `no-linker` feature is _not_ enabled
_and_ one is not on a platform where that implementation is selected
regardless of the feature. One obvious case is illumos.

Unfortunately, platform-specific features are not yet implemented. This
commit instead makes the `dof` crate an optional dependency on macOS
platforms, and a required dependency on all other platforms. Note that
the `dof` crate is still pulled in on macOS if the `no-linker` feature
is explicitly enabled.

This commit also removes the `--features no-linker` bits from the
buildomat script, and bumps versions in preparation for a release with
this new dependency graph.

### DIFF
--- a/usdt-tests-common/src/lib.rs
+++ b/usdt-tests-common/src/lib.rs
@@ -1,11 +1,10 @@
-use std::env;
 use std::process::Command;
 
 #[cfg(target_os = "illumos")]
 pub fn root_command(name: &str) -> Command {
     // On illumos systems, we prefer pfexec(1) but allow some other command to
     // be specified through the environment.
-    let pfexec = env::var("PFEXEC").unwrap_or_else(|_| "/usr/bin/pfexec".to_string());
+    let pfexec = std::env::var("PFEXEC").unwrap_or_else(|_| "/usr/bin/pfexec".to_string());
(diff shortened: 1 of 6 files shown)


usdt-tests-common/src/lib.rs
```diff
@@ -1,11 +1,10 @@
-use std::env;
 use std::process::Command;
 
 #[cfg(target_os = "illumos")]
 pub fn root_command(name: &str) -> Command {
     // On illumos systems, we prefer pfexec(1) but allow some other command to
     // be specified through the environment.
-    let pfexec = env::var("PFEXEC").unwrap_or_else(|_| "/usr/bin/pfexec".to_string());
+    let pfexec = std::env::var("PFEXEC").unwrap_or_else(|_| "/usr/bin/pfexec".to_string());
```
jclulow, line 7: So that I know in future: is there some rule/heuristic you're using to determine whether to import a module or merely fully qualify the call site?

jclulow (approved):
I believe this makes sense!  Thanks for fixing this.

PR #28, opened 2021-10-12
Improves and simplifies feature selection

bnaecker:
- Removes the `no-linker` feature, which was mainly used during
  development to artificially select the "no-linker" implementation for
  testing and development on platforms where that is not the default
  (e.g., macOS).
- Adds the ability to select the empty or no-op implementation in the
  whole package. This is done with the `--no-default-features` flag or
  `default-features = false` key in a Cargo.toml. This selects the
  empty.rs implementation, regardless of platform. Other than this, the
  implementation is selected on the basis of platform entirely; there's
  no longer a way to choose between the linker/no-linker versions on the
  command-line. This relies on our new illumos CI system using
  buildomat.
- Removes the unnecessary "asm" feature from the macro crates. These are
  always the same, regardless of this feature.
- Simplifies GitHub CI. The unit tests are now run on macOS and Ubuntu,
  for both the empty and platform-default implementation. Integration
  tests are now only run on macOS.
- Adds a `buildomat` test verifying the empty implementation passes
  _unit_ tests, using a stable compiler. This verifies that the
  integration tests and examples compile correctly, but does not run
  them.


README.md
```diff
@@ -222,6 +223,45 @@ comes with significant tradeoffs. As such the current recommendation is:
...
+[features]
+usdt-probes = ["usdt/asm"]
+```
+
+This allows users to opt into probes if they're willing to accept a nightly toolchain.
+
+### The Rust `asm` feature
+
+Recall from the example that the `usdt` crate relies on inline `asm`, which is not yet a
+stable Rust feature. This means that the code _calling_ the generated probe macros must
```
ahl, line 254: in case you want to drop this in there: https://github.com/rust-lang/rust/issues/72016

ahl (commented):
ship it

PR #29, opened 2021-10-15
Fixes type checking for probes with zero arguments

bnaecker:
The internal type-checking mechanisms were broken for probes with no
arguments. One could write an argument closure returning _any_
arguments, and this would "pass" our internal type-checking, as we only
unpacked the _expected_ arguments from the closure. As there were no
expected arguments, there was no type-checking.

This commit modifies type-checking to verify that if the probe expects
zero arguments, we can assign the result of calling the argument closure
to a variable of type `()`. It also modifies the type checking of
serializable types. Before we used a generic parameter with a trait
bound `T: ::serde::Serialize`. This is actually not enough, since it
would successfully type-check an argument closure that returns a
_different_ type that is serializable than the one in the probe
signature. We now collect the actual, concrete type from the probe
signature during parsing, and insert that into the signature of the
type-check function as we would any other concrete type.

Two new try-build tests have been added that verify this behavior.

### DIFF
--- a/tests/compile-errors/src/different-serializable-type.stderr
+++ b/tests/compile-errors/src/different-serializable-type.stderr
@@ -1,0 +1,10 @@
+error[E0308]: mismatched types
+  --> $DIR/different-serializable-type.rs:18:1
+   |
+18 | #[usdt::provider]
+   | ^^^^^^^^^^^^^^^^^ expected struct `Expected`, found struct `Different`
+...
+25 |     my_provider_my_probe!(|| Different { x: 0 });
+   |     --------------------------------------------- in this macro invocation
+   |
+   = note: this error originates in the macro `my_provider_my_probe` (in Nightly builds, run with -Z macro-backtrace for more info)
--- a/tests/compile-errors/src/lib.rs
+++ b/tests/compile-errors/src/lib.rs
@@ -10,5 +10,7 @@ mod tests {
         t.compile_fail("src/unsupported-type.rs");
         t.compile_fail("src/no-closure.rs");
         t.compile_fail("src/no-provider-file.rs");
+        t.compile_fail("src/zero-arg-probe-type-check.rs");
+        t.compile_fail("src/different-serializable-type.rs");
     }
 }
--- a/usdt-impl/src/common.rs
+++ b/usdt-impl/src/common.rs
@@ -8,9 +8,26 @@ use crate::DataType;
 
 // Construct function call that is used internally in the UDST-generated macros, to allow
 // compile-time type checking of the lambda arguments.
-pub fn generate_type_check(types: &[DataType]) -> TokenStream {
+pub fn generate_type_check(
+    provider_name: &str,
+    probe_name: &str,
+    types: &[DataType],
+) -> TokenStream {
+    // If the probe has zero arguments, verify that the result of calling the closure is `()`
+    if types.is_empty() {
+        return quote! {
+            {
+                let _ = || {
+                    let _: () = $args_lambda();
+                };
+            }
+        };
+    }
+
+    // For one or more arguments, verify that we can unpack the closure into a tuple of type
+    // `(arg0, arg1, ...)` and those items can be passed to a function with the same signature as
+    // the probe, i.e., `fn(type0, type1, ...)`.
     let mut has_strings = false;
-    let mut has_serializables = false;
     let type_check_args = types
         .iter()
         .map(|typ| match typ {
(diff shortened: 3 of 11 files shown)


tests/compile-errors/src/different-serializable-type.stderr
```diff
@@ -0,0 +1,10 @@
+error[E0308]: mismatched types
+  --> $DIR/different-serializable-type.rs:18:1
+   |
+18 | #[usdt::provider]
+   | ^^^^^^^^^^^^^^^^^ expected struct `Expected`, found struct `Different`
+...
+25 |     my_provider_my_probe!(|| Different { x: 0 });
```
ahl, line 7: I'm not sure how tenable it is, but it would be cool if this underlined `Different { x: 0 }`

tests/compile-errors/src/lib.rs
```diff
@@ -10,5 +10,7 @@ mod tests {
         t.compile_fail("src/unsupported-type.rs");
         t.compile_fail("src/no-closure.rs");
         t.compile_fail("src/no-provider-file.rs");
+        t.compile_fail("src/zero-arg-probe-type-check.rs");
+        t.compile_fail("src/different-serializable-type.rs");
```
ahl, line 14: What do you think of having a directory of these tests as in dropshot?

https://github.com/oxidecomputer/dropshot/blob/main/dropshot/tests/fail.rs

usdt-impl/src/common.rs
```diff
@@ -8,26 +8,45 @@ use crate::DataType;
...
 // compile-time type checking of the lambda arguments.
-pub fn generate_type_check(types: &[DataType]) -> TokenStream {
+pub fn generate_type_check(
+    provider_name: &str,
+    probe_name: &str,
+    types: &[DataType],
+) -> TokenStream {
+    // If the probe has zero arguments, verify that the result of calling the closure is `()`
+    if types.is_empty() {
+        return quote! {
```
ahl, line 18: I think if we passed in the closure object or the span for the closure we could use `quote_spanned!` to have the error point at the closure more specifically.

usdt-impl/src/common.rs
```diff
@@ -37,23 +56,37 @@ pub fn generate_type_check(types: &[DataType]) -> TokenStream {
...
+        quote! {}
     };
 
-    // NOTE: This block defines an internal empty function and then a lambda which
-    // calls it. This is all strictly for type-checking, and is optimized out. It is
-    // defined in a scope to avoid multiple-definition errors in the scope of the macro
-    // expansion site.
+    let type_check_function =
+        format_ident!("__usdt_private_{}_{}_type_check", provider_name, probe_name);
     quote! {
```
ahl, line 84: again, I think if we put this all under `quote_spanned!` it might have a slightly better ui for the error message.

usdt-impl/src/no-linker.rs
```diff
@@ -60,27 +60,29 @@ fn compile_probe(
...
-                "990:   clr rax",
-                #is_enabled_rec,
-                out("rax") is_enabled,
-                options(nomem, nostack, preserves_flags)
-            );
-        }
-
-        if is_enabled != 0 {
-            #unpacked_args
+        {
```
ahl, line 63: this was important to keep the `is_enabled` variable out of the namespace, correct?

usdt-impl/src/no-linker.rs
```diff
@@ -60,27 +60,29 @@ fn compile_probe(
...
+            let mut is_enabled: u64;
+            // TODO can this block be option(pure)?
```
ahl, line 65: Neither here nor there, but it seems like the answer to this question is: definitely not. Like the last thing we want to be doing is giving the compiler an invitation to poke inside this.

ahl (commented):
ship it! some suggestions about improved failure ui stuff, but those will 1. probably take some monkeying around to get working and 2. can be safely deferred

usdt-impl/src/no-linker.rs
bnaecker: Yeah, exactly, I want to pollute the caller's namespace as little as possible.

tests/compile-errors/src/lib.rs
bnaecker: I think that's fine, but I'm going to defer this one.

usdt-impl/src/common.rs
ahl: as discussed: I was wrong about its applicability here in that we're generating macros and the errors come from those macros

bnaecker: Ok, so I think this won't apply here. We're generating a macro_rules! macro from this code, and we don't have the span associated with the argument closure that the caller ultimately passes in here. We'll have to make do with this for now, so I'll resolve this and the related comments. Thanks for the review!

tests/compile-errors/src/different-serializable-type.stderr
bnaecker: Also deferring this one. It'd be nice, but I don't think we have any way to get the span of the closure from our proc macros.

PR #32, opened 2021-10-20
Add default features to usdt-impl crate

bnaecker:
This fixes an annoying issue related to feature-detection. Specifically,
it seems that we need the `usdt-impl` crate to specify the `asm` feature
as a default for the `default-features = false` bit in the top-level
`usdt` crate to work. Without this, it's not clear why, but the
`usdt-impl/asm` feature may or may not be selected dependending on how
consumers of the `usdt` crate set their feature flags.

This also bumps the version numbers, in preparation for a release.

### DIFF
--- a/usdt-impl/src/record.rs
+++ b/usdt-impl/src/record.rs
@@ -164,6 +164,19 @@ pub(crate) fn addr_to_info(addr: u64) -> (Option<String>, Option<String>) {
     }
 }
 
+// Limit a string to the DTrace-imposed maxima. Note that this ensures a null-terminated C string
+// result, i.e., the actula string is of length `limit - 1`.
+// See dtrace.h
+const MAX_PROVIDER_NAME_LEN: usize = 64;
+const MAX_PROBE_NAME_LEN: usize = 64;
+const MAX_FUNC_NAME_LEN: usize = 128;
+const MAX_ARG_TYPE_LEN: usize = 128;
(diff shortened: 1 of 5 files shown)


usdt-impl/src/record.rs
```diff
@@ -186,32 +199,31 @@ fn process_rec(providers: &mut BTreeMap<String, Provider>, rec: &[u8]) -> Result
...
 
     let funcname = match addr_to_info(address).0 {
-        Some(s) => s,
+        Some(s) => limit_string_length(s, MAX_FUNC_NAME_LEN),
         None => format!("?{:#x}", address),
     };
 
-    let provider = providers.entry(provname.to_string()).or_insert(Provider {
-        name: provname.to_string(),
+    let provname = limit_string_length(provname, MAX_PROVIDER_NAME_LEN);
```
ahl, line 212: ok... this is picayune, but if we're going to check this we need to keep in mind that DTrace will append the pid to the process name specified so it's actually MAX_PROVIDER_NAME_LEN - log_10(maxpid) and by default that's 5 or 6

ahl (commented):
good stuff; nice work fixing this

usdt-impl/src/record.rs
ahl: Yes of course. And no need to apologize as it's totally opaque and basically impossible for you to have discovered.

PR #35, opened 2021-11-01
Always borrow probe data when possible

bnaecker:
This PR does a few things.

- Data passed to a probe macro is always borrowed. This uses the `std::borrow::Borrow` trait when possible, and sometimes the `AsRef` trait where appropriate (slices, strings).
- Updates the type-checks inside the generate probe macros to first clone the closure. Neither the closure nor the captured data are ever actually cloned, since the closure is used inside a closure that's never called. Cloning is just to avoid compiler errors complaining about using data after it's moved.
- Adds the `Clone` trait bound to serializable types. This should be a minimal burden on users, and supports the previous point.
- Updates some documentation


bnaecker:
Discussed offline with @ahl, merging.

PR #36, opened 2021-11-02
Implements one-time registration of probes in a process

bnaecker:
This commit adds support for guaranteed at-most-once registration of
probes, within a single process lifetime. The linker section into which
the probes are written is now writable. When reading a probe record, the
code checks that the probe's version is supported, and, if so,
atomically exchanges it for a sentinel value (255), returning the
original. Since any probe version greater than the crate's version is
skipped, probes that are handled will be skipped in any other invocation
of the `register_probes` function, as those probes' version is now the
sentinel. Unhandled probes will not be modified, but will still be
skipped, letting later versions of the library handle them.

### DIFF
--- a/usdt-impl/src/record.rs
+++ b/usdt-impl/src/record.rs
@@ -1,5 +1,6 @@
 //! Implementation of construction and extraction of custom linker section records used to store
 //! probe information in an object file.
+
 // Copyright 2021 Oxide Computer Company
 
 use std::{
@@ -18,6 +19,12 @@ use libc::{c_void, Dl_info};
 #[cfg(feature = "des")]
 use goblin::Object;
 
+use crate::DataType;
+
+// Version number for probe records containing data about all probes.
+//
+// NOTE: This must have a maximum of `u8::MAX - 1`. See `read_and_update_record_version` for
+// details.
 pub(crate) const PROBE_REC_VERSION: u8 = 1;
 
 /// Extract probe records from the given file, if possible.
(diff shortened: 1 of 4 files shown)


usdt-impl/src/record.rs
```diff
@@ -169,16 +176,53 @@ fn limit_string_length<S: AsRef<str>>(s: S, limit: usize) -> String {
...
+        // NOTE: It's not easy to use types from `std::sync::atomic` here because we need to
+        // atomically exchange the data through a pointer. `AtomicU8::from_mut` might work, but
+        // that would require another feature flag pinning us to a nightly compiler.
+        let mut version = u8::MAX;
+        let mut existing_version = data.as_ptr();
+        #[allow(unused_assignments)]
+        unsafe {
+            asm!(
+                "lock xchg al, [{}]",
+                inout(reg) existing_version,
```
ahl, line 196: is this `inout` or just `in`?

usdt-impl/src/record.rs
```diff
@@ -169,16 +176,53 @@ fn limit_string_length<S: AsRef<str>>(s: S, limit: usize) -> String {
...
+    // or equal to the crate's version.
+    if data[0] <= PROBE_REC_VERSION {
+        // Atomically exchange the record's version with our sentinel value, and return the
+        // record's version.
+        //
+        // NOTE: It's not easy to use types from `std::sync::atomic` here because we need to
+        // atomically exchange the data through a pointer. `AtomicU8::from_mut` might work, but
+        // that would require another feature flag pinning us to a nightly compiler.
+        let mut version = u8::MAX;
+        let mut existing_version = data.as_ptr();
```
ahl, line 191: might this more accurately be named something like ... record_version_ptr?

bnaecker: Good idea, fixed in bc48dc9

usdt-impl/src/record.rs
bnaecker: Nope, nice catch. Fixed and removed the `mut` annotation in bc48dc9

PR #44, opened 2021-11-09
Supports crates using the new resolver v2

bnaecker:
The dependency resolver v2 does not merge features in the same way as
prior versions. V2 is also the default as of Rust edition 2021. This
could lead to incorrectly building multiple versions of the usdt crate,
including one erroneously selecting the empty implementation.

This adds feature flags for the `usdt-macro` and `usdt-attr-macro`
crates that ensure we're building with the correct set of flags
regardless of which resolver is used.

### DIFF
--- a/usdt-attr-macro/Cargo.toml
+++ b/usdt-attr-macro/Cargo.toml
@@ -18,5 +18,9 @@ syn = { version = "1", features = ["full"] }
 quote = "1"
 usdt-impl = { path = "../usdt-impl", version = "0.1.14", default-features = false }
 
+[features]
+default = ["asm"]
+asm = ["usdt-impl/asm"]
+
 [dev-dependencies]
 rstest = "0.11"
--- a/usdt-macro/Cargo.toml
+++ b/usdt-macro/Cargo.toml
@@ -16,5 +16,9 @@ syn = { version = "1", features = ["full"] }
 quote = "1"
 usdt-impl = { path = "../usdt-impl", version = "0.1.14", default-features = false }
 
+[features]
+default = ["asm"]
+asm = ["usdt-impl/asm"]
+
 [lib]
 proc-macro = true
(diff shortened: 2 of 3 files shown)


bnaecker:
As an example of the effect of this, I created a small test binary depending on `usdt` by local path. This uses `edition = "2018"` in its Cargo.toml file. Running `cargo +nightly fix --edition` shows changes in dependency resolution, especially related to feature flags, that would result from updating to the new resolver or `edition = "2021"`. The result, on the current version of `master`, is:

```bash
$ cargo +nightly fix --edition
note: Switching to Edition 2021 will enable the use of the version 2 feature resolver in Cargo.
This may cause some dependencies to be built with fewer features enabled than previously.
More information about the resolver changes may be found at https://doc.rust-lang.org/nightly/edition-guide/rust-2021/default-cargo-resolver.html
When building the following dependencies, the given features will no longer be used:

  syn v1.0.81 removed features: visit
  syn v1.0.81 (as host dependency) removed features: visit
  usdt-impl v0.1.14 (/Users/bnaecker/file-cabinet/oxide/usdt/usdt-impl) (as host dependency) removed features: asm
...
```

We can see that the new resolution would remove the feature `asm` from the `usdt-impl` crate, which is kinda the whole point of this package. I used `cargo tree -e features` to see where this was coming from, the result of which is that the macro crates were not correctly propagating the features to `usdt-impl`. This PR fixes that. Running the same `cargo fix` command on this commit results in:

```bash
$ cargo +nightly fix --edition
note: Switching to Edition 2021 will enable the use of the version 2 feature resolver in Cargo.
This may cause some dependencies to be built with fewer features enabled than previously.
More information about the resolver changes may be found at https://doc.rust-lang.org/nightly/edition-guide/rust-2021/default-cargo-resolver.html
When building the following dependencies, the given features will no longer be used:

  syn v1.0.81 removed features: visit
  syn v1.0.81 (as host dependency) removed features: visit
```

I.e. the `asm` feature is now propagated correctly.

Issue #41, opened 2021-11-08
Add flag for new `asm_sym` feature


bnaecker:
The tests for PR #40 originally failed. This is unrelated to the code in the PR itself, but due to a [recently-merged set of feature flags](https://github.com/rust-lang/rust/pull/90348) related to inline assembly.

USDT uses inline assembly in its implementation, which has been in the stabilization process for a while. That's behind a feature flag `asm`, but during stabilization, that feature has been broken out into several more fine-grained ones. This specific message is due to the use of the [symbol operands](https://doc.rust-lang.org/unstable-book/library-features/asm.html#symbol-operands-and-abi-clobbers) on macOS.

bnaecker:
Closed by #45.

PR #45, opened 2021-11-09
Adds documentation about new `asm_sym` feature

bnaecker:
This commit adds documentation and fixes tests for the new `asm_sym`
feature, which gates symbol operands in inline assembly. These operands
are integral to this crate's implementation on macOS. Since we're using
a nightly toolchain, this is a breaking change, and there's no way to
build `usdt` or downstream crates that compile both before and after the
addition of this feature. The documentation describes this clearly, and
provides the feature flags needed on either side of this divide.

The tests have also been updated with the feature flag, as they run on
the latest nightly.


bnaecker:
This will address #41 when merged.

probe-test-attr/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2018"
 [dependencies]
 usdt = { path = "../usdt", default-features = false }
 serde = "1"
+rustversion = "1"
```
ahl, line 10: stale?

bnaecker: Yep, thanks! Fixed in 84a6eac.

PR #52, opened 2022-02-25
Wire up support for stable asm!()

pfmooney:
Support for `asm!()` has finally landed in stable rust!  This should make the crate usable by those building with 1.59 stable and later.  Earlier nightly compilers should continue to work, provided the `feature(asm)` declaration is still in the consumer crate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,9 @@ than the first two. See [below](#serializable-types) for more details, but brief
 form supports probe arguments of any type that implement [`serde::Seralize`][2]. These different
 versions are shown in the crates `probe-test-{build,macro,attr}` respectively.
 
-> Note: This crate uses inline assembly to work its magic. As such a nightly Rust toolchain is
-required, and the functionality is hidden behind the `"asm"` feature flag. A nightly toolchain
-can be installed with `rustup toolchain install nightly`. See [the notes](#notes) for a
-discussion.
+> Note: This crate uses inline assembly to work its magic.  That feature has been stable as of Rust
+1.59.  Any project using `usdt` prior to that must build with a nightly toolchain and enable the
+`asm` feature.  See the [notes](#notes) for a discussion.
 
 ## Example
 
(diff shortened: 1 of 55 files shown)


ahl:
This is awesome! Thank you. Can you update the README.md as well? I would suggest writing it from the perspective of 1.59+ being the norm, but perhaps with a section devoted to pre 1.59

This "test_does_it_work" looks potentially unstable... you might check with @bnaecker if it's not an easy one to resolve.

pfmooney:
> Can you update the README.md as well?

I've tweaked it to emphasize the inline-asm-is-now-stable case.  I'm happy to defer if you think it should be done differently.

> This "test_does_it_work" looks potentially unstable... you might check with @bnaecker if it's not an easy one to resolve.

I'm not sure what's going on there.  On my illumos (omnios) machine, the `does-it-work` test passes, but the `relative-import` test fails (due to compiler error output being slightly different between nightly and stable).

README.md
```diff
@@ -232,11 +231,12 @@ guarantee that probes are registered.
 
 ## Notes
 
-The `usdt` crate requires a nightly toolchain, as it relies on the currently-unstable [inline
-asm][inline-asm] feature. However, the crate contains an empty, no-op implementation, which
-generates all the same probe macros, but with empty bodies. This may be selected by passing
-the `--no-default-features` flag when building the crate, or by using `default-features = false`
-in the [`[dependencies]` table][feature-deps] of one's `Cargo.toml`.
+The `usdt` crate requires [inline asm][inline-asm], a feature only stabilized as of Rust 1.59.
+Prior to that version, a nightly toolchain was required to to import the feature.  For legacy
```
bnaecker, line 235: Typo: "to to"

README.md
```diff
@@ -253,14 +253,15 @@ usdt = { version = "*", optional = true, default-features = false }
...
 
 ### The Rust `asm` feature
 
-Recall from the example that the `usdt` crate relies on inline `asm`, which is [not yet][asm-issue] a
-stable Rust feature. This means that the code _calling_ the generated probe macros must
-be in a module where the `asm!` macro can be used, i.e., where the `feature(asm)` configuration
-directive is applicable.
+On toolchains prior to 1.59, inline asm was [not available][asm-issue] without the feature being
+enabled.  This applies to code _calling_ the probe macros, in addition to `usdt` where they are
+implemented.  Those generated probe macros must be in a module that is either built with a >=1.59
```
bnaecker, line 263: I'd be explicit here: ">= 1.59 stable toolchain or a nightly toolchain where the `feature(asm)`..."

bnaecker: Actually, I think there should be a note here about macOS. Something pointing out that macOS still requires nightly, due to the `asm_sym` and `asm_const` features not yet on stable.

bnaecker (approved):
Thanks for taking the lead on this! A couple of README comments, but otherwise looks great.

ahl:
I think the `test_compile_errors` failure may be due merely to a change in the rustc error formatting..

bnaecker:
@ahl Yeah, agreed. I'm not sure those tests are valuable anymore, to be honest. They were useful during development, to make sure we had helpful messages. But now that we do, it seems only a test of the internals of these messages. And given that we'll be supporting both stable and nightly (for macOS) for a while, I doubt those messages will be consistent enough to test.

ahl (approved):

PR #56, opened 2022-04-01
Add support for const pointers to integer types

bnaecker:
- Reworks some of the internal dtrace-parser code to make integers more
  compact and easily represent pointers.
- Adds support for pointers the implementation crates
- Adds pointers to list of supported types
- Adds a probe in the `probe-test-attr` crate that uses a pointer.

### DIFF
--- a/dtrace-parser/src/dtrace.pest
+++ b/dtrace-parser/src/dtrace.pest
@@ -17,9 +17,8 @@ IDENTIFIER = @{ ASCII_ALPHA+ ~ (ASCII_ALPHANUMERIC | "_")* }
 BIT_WIDTH = @{ "8" | "16" | "32" | "64" }
 SIGNED_INT = ${ "int" ~ BIT_WIDTH ~ "_t" }
 UNSIGNED_INT = ${ "uint" ~ BIT_WIDTH ~ "_t" }
-INT = $ { SIGNED_INT | UNSIGNED_INT }
 STAR = ${ "*" }
-INTEGER = ${ (INT ~ STAR) | INT }
+INTEGER = ${ ((SIGNED_INT | UNSIGNED_INT) ~ STAR) | (SIGNED_INT | UNSIGNED_INT) }
 STRING = { "char" ~ STAR }
 DATA_TYPE = { INTEGER | STRING }
 
--- a/dtrace-parser/src/dtrace.pest
+++ b/dtrace-parser/src/dtrace.pest
@@ -1,5 +1,5 @@
 // A definition of the basic grammar of DTrace provider definitions.
-// Copyright 2021 Oxide Computer Company
+// Copyright 2022 Oxide Computer Company
 
 // Some basic tokens
 PROBE_KEY = @{ "probe" }
@@ -17,8 +17,11 @@ IDENTIFIER = @{ ASCII_ALPHA+ ~ (ASCII_ALPHANUMERIC | "_")* }
 BIT_WIDTH = @{ "8" | "16" | "32" | "64" }
 SIGNED_INT = ${ "int" ~ BIT_WIDTH ~ "_t" }
 UNSIGNED_INT = ${ "uint" ~ BIT_WIDTH ~ "_t" }
-STRING = { "char" ~ "*" }
-DATA_TYPE = { STRING | UNSIGNED_INT | SIGNED_INT}
+INT = $ { SIGNED_INT | UNSIGNED_INT }
+STAR = ${ "*" }
+INTEGER = ${ (INT ~ STAR) | INT }
+STRING = { "char" ~ STAR }
+DATA_TYPE = { INTEGER | STRING }
 
 // A list of probe arguments, which are just data types
 ARGUMENT_LIST = { ( DATA_TYPE ~ ("," ~ DATA_TYPE)* )* }
--- a/dtrace-parser/src/dtrace.pest
+++ b/dtrace-parser/src/dtrace.pest
@@ -17,10 +17,11 @@ IDENTIFIER = @{ ASCII_ALPHA+ ~ (ASCII_ALPHANUMERIC | "_")* }
 BIT_WIDTH = @{ "8" | "16" | "32" | "64" }
 SIGNED_INT = ${ "int" ~ BIT_WIDTH ~ "_t" }
 UNSIGNED_INT = ${ "uint" ~ BIT_WIDTH ~ "_t" }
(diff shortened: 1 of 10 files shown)


dtrace-parser/src/dtrace.pest
```diff
@@ -17,8 +17,11 @@ IDENTIFIER = @{ ASCII_ALPHA+ ~ (ASCII_ALPHANUMERIC | "_")* }
 BIT_WIDTH = @{ "8" | "16" | "32" | "64" }
 SIGNED_INT = ${ "int" ~ BIT_WIDTH ~ "_t" }
 UNSIGNED_INT = ${ "uint" ~ BIT_WIDTH ~ "_t" }
-STRING = { "char" ~ "*" }
-DATA_TYPE = { STRING | UNSIGNED_INT | SIGNED_INT}
+INT = $ { SIGNED_INT | UNSIGNED_INT }
+STAR = ${ "*" }
+INTEGER = ${ (INT ~ STAR) | INT }
```
jgallagher, line 22: This is super nitpicky, but having both `INT` and `INTEGER` (that aren't synonymous) seems a little dubious. Names are hard; maybe something like `INTEGER` (for what is currently `INT`) and `INTEGER_TYPE` (for int or int pointer)? I don't feel strongly about this; feel free to ignore.

usdt-attr-macro/src/lib.rs
```diff
@@ -209,7 +237,10 @@ fn parse_probe_argument(
         }
         _ => Err(syn::Error::new(
             item.span(),
-            "Probe arguments must be path types, slices, arrays, tuples or references",
+            concat!(
```
jgallagher, line 240: Ooh, nice, I never thought of using `concat!` to preserve column width. :+1:

usdt-attr-macro/src/lib.rs
```diff
@@ -209,7 +237,10 @@ fn parse_probe_argument(
...
+                "Probe arguments must be path types, slices, arrays, tuples, ",
+                "references, or const pointers to integers ",
```
jgallagher, line 242:
```suggestion
                "references, or const pointers to integers",
```

jgallagher (approved):
LGTM! Tried it with on MGS locally and worked perfectly.

dtrace-parser/src/dtrace.pest
bnaecker: Yeah, that's a bit awkward, I agree. I might just get rid of `INT` entirely, and use something like `(SIGNED_INT | UNSIGNED_INT) ~ STAR` as the rule instead. Pest has been helpful, but it's really unwieldy. I'll experiment!

jgallagher: Yeah, that looks good :+1:

dtrace-parser/src/dtrace.pest
```diff
@@ -17,8 +17,10 @@ IDENTIFIER = @{ ASCII_ALPHA+ ~ (ASCII_ALPHANUMERIC | "_")* }
 BIT_WIDTH = @{ "8" | "16" | "32" | "64" }
 SIGNED_INT = ${ "int" ~ BIT_WIDTH ~ "_t" }
 UNSIGNED_INT = ${ "uint" ~ BIT_WIDTH ~ "_t" }
-STRING = { "char" ~ "*" }
-DATA_TYPE = { STRING | UNSIGNED_INT | SIGNED_INT}
+STAR = ${ "*" }
+INTEGER = ${ ((SIGNED_INT | UNSIGNED_INT) ~ STAR) | (SIGNED_INT | UNSIGNED_INT) }
```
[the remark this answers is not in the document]
ahl: or maybe even

```
INTEGER = ${ (SIGNED_INT | UNSIGNED_INT | "char") }
```

usdt/src/lib.rs
```diff
@@ -211,6 +212,7 @@
 //! Below is the full list of supported types.
 //!
 //! - `(u?)int(8|16|32|64)_t`
```
ahl, line 214: huh. interesting that we can't do `char` here (see comments in the parser)

ahl (approved):
let's get that money!

dtrace-parser/src/dtrace.pest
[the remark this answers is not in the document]
bnaecker: This was my initial approach. It's probably better, but led me to a lot more code in Rust. That was probably because I tried to represent the pointer-to-int as another enum variant, transliterating the grammar. I'll give it another whirl, though, see if I can't make it simpler.

usdt/src/lib.rs
bnaecker: We could, but I'm not sure how that'd play with handling actual Rust strings. We're kind of reserving `char*` to mean a string, so I'd want to make sure we don't conflate that with a byte array.

dtrace-parser/src/dtrace.pest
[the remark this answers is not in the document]
bnaecker: Ok, I've made the data type rule effectively what you've got here, and also have a separate enum in the Rust representation for a pointer-to-int. The pointer-ness of the type was previously a boolean field. Let me know if this seems better!

usdt/src/lib.rs
ahl: makes sense..

dtrace-parser/src/dtrace.pest
```diff
@@ -17,10 +17,11 @@ IDENTIFIER = @{ ASCII_ALPHA+ ~ (ASCII_ALPHANUMERIC | "_")* }
 BIT_WIDTH = @{ "8" | "16" | "32" | "64" }
 SIGNED_INT = ${ "int" ~ BIT_WIDTH ~ "_t" }
 UNSIGNED_INT = ${ "uint" ~ BIT_WIDTH ~ "_t" }
+INTEGER = ${ (SIGNED_INT | UNSIGNED_INT) }
```
ahl, line 20: just observing that we can't pass `char` as a parameter type... which is probably fine

dtrace-parser/src/dtrace.pest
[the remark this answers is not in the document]
ahl: I think it maybe looks better; but I defer to you. I didn't realize how it might impact the representation; I hope I didn't send you on the wrong path..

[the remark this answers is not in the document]
bnaecker: Not at all. I had tried it originally, and flaked out due to laziness I think. I like this better, considering integers and pointers different types.

PR #60, opened 2022-04-21
More stable asm

pfmooney:
This retools `asm` and `asm_sym` feature inside the usdt implementation to make fallback more smooth when built with a compiler which does not support the necessary features.  Consumers wanting the old behavior, where the crate will fail to build if they do not have a supported compiler and necessary enabled features, is preserved with the `strict-asm` feature.  Without `strict-asm` enabled, the usdt-impl `build.rs` will quietly fall back to the no-op probe implementation if compiler functionality is not met.

### DIFF
--- a/common-build.rs
+++ b/common-build.rs
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-use version_check;
-
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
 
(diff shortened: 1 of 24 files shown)


bnaecker (approved):
Awesome, thanks for taking this on!

ahl:
This looks great. Thanks for doing it. Am I right in inferring that this is a bunch of cleanup, but that the meat of it is [usdt-impl/build.rs](https://github.com/oxidecomputer/usdt/pull/60/files#diff-023323405774884369480a2a63eb32532f111eef0ae73dcb29ef35ad6a577c09)?

pfmooney:
> This looks great. Thanks for doing it. Am I right in inferring that this is a bunch of cleanup, but that the meat of it is [usdt-impl/build.rs](https://github.com/oxidecomputer/usdt/pull/60/files#diff-023323405774884369480a2a63eb32532f111eef0ae73dcb29ef35ad6a577c09)?

Aye. The substantial change around `asm` features is in eac0fe5f03c3fbf23468ead5cb140f62d51ac3f3.

PR #67, opened 2022-11-18
Bump versions for release, add pinned-toolchain for err checks

bnaecker:


ahl:
I'm fine with whatever you choose, but I'd suggest we keep the compile-time output checks to keep tabs on the current rustc output.

bnaecker:
Thanks Adam! Just checking, but do you mean keep the output and the tests themselves, but don't necessarily run them in CI?

ahl:
In dropshot we pin the rust toolchain https://github.com/oxidecomputer/dropshot/blob/main/rust-toolchain.toml and use renovate to nudge that forward. I'd advocate that we run the build tests in CI, but the only times we should see these (possibly) spurious changes is when we rev the rust version.

(did you mean to add the rust-toolchain?)

bnaecker:
Yeah, I had already pinned the toolchain in the Buildomat job description, yet still saw a failure in those tests. Of course I can't find the output now. I'll try it again.

bnaecker:
Ok, so [this](https://buildomat.eng.oxide.computer/wg/0/details/01GJ6CKASGTRRWQS2QEK4VMA0E/QcrPvW8UlR5zqaSyW2HAAEzo6f5ogQFrCm4fpuwcoiiO5mvt/01GJ6CKK4GT8F4474ANAWD4RZR) is what I was talking about. That is the Buildomat output of a run just checking the compiler error messages. It uses a toolchain pinned to `nightly-2021-11-24`. Running that on my Helios box completes successfully, but it fails in CI with a single different character AFAICT (a spurious `LF` in the `pest` error message). I have no idea why that is, these should be running exactly the same code with exactly the same toolchain.

bnaecker:
Ugh, lockfiles! We need to prevent the lockfile from being updated in CI, this was pulling a newer version of `pest`, which changed the error outputs.

PR #69, opened 2022-11-22
Add both path and version dependencies

bnaecker:
A previous commit removed the version specification for the path dependencies in the repository. This all works fine, until the crates are published. Crates on crates.io may not have path dependencies, but there is no version to use.

The accepted solution appears to be to specify _both_ path and version dependencies. The former is used when building from source, the latter when publishing.

### DIFF
--- a/usdt/Cargo.toml
+++ b/usdt/Cargo.toml
@@ -8,12 +8,19 @@ license = "Apache-2.0"
 description = "Dust your Rust with USDT probes"
 repository = "https://github.com/oxidecomputer/usdt.git"
 
+# NOTE: The use of path and version dependencies is deliberate and load-bearing.
+#
+# When building from source, this will select the version in the workspace, by
+# path. When publishing the specific version will be used. Combined with the
+# exact version requirements, this prevents issues like oxidecomputer/usdt#69,
+# where the public-facing `usdt` crate may pick up different versions of the
+# internal implementation crates.
 [dependencies]
-dtrace-parser = { path = "../dtrace-parser", optional = true, version = "0.1.14" }
+dtrace-parser = { path = "../dtrace-parser", optional = true, version = "= 0.1.14" }
 serde = "1"
-usdt-impl = { path = "../usdt-impl", default-features = false, version = "0.3.4" }
-usdt-macro = { path = "../usdt-macro", default-features = false, version = "0.3.4" }
-usdt-attr-macro = { path = "../usdt-attr-macro", default-features = false, version = "0.3.4" }
+usdt-impl = { path = "../usdt-impl", default-features = false, version = "= 0.3.4" }
+usdt-macro = { path = "../usdt-macro", default-features = false, version = "= 0.3.4" }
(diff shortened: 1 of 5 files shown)


usdt/Cargo.toml
```diff
@@ -9,11 +9,11 @@ description = "Dust your Rust with USDT probes"
 repository = "https://github.com/oxidecomputer/usdt.git"
 
 [dependencies]
```
gjcolombo, line 11: Worth throwing a comment on here noting that the version specs are load-bearing for publishing purposes?

gjcolombo (approved):
LGTM with one small comment.

I did a little browsing, and it looks like `cargo package` can help detect the "no version information" case; if I run it against `master` I get errors of the following form:

```
error: all dependencies must have a version specified when packaging.                                  
dependency `dtrace-parser` does not specify a version                                                                                   
Note: The packaged dependency will use the version from crates.io,                                                                      
the `path` specification will be removed from the dependency declaration. 
```

If I run it against the published packages in this branch, I get different errors about unsatisfiable dependencies, but I think that's expected and can be resolved by publishing things in the right order. (For example, `usdt-macro` complains about not being able to find `dtrace-parser` 0.1.14, which makes sense since it's not published yet.)

usdt/Cargo.toml
```diff
@@ -9,11 +9,11 @@ description = "Dust your Rust with USDT probes"
...
-dtrace-parser = { path = "../dtrace-parser", optional = true }
+dtrace-parser = { path = "../dtrace-parser", optional = true, version = "0.1.14" }
 serde = "1"
-usdt-impl = { path = "../usdt-impl", default-features = false }
-usdt-macro = { path = "../usdt-macro", default-features = false }
-usdt-attr-macro = { path = "../usdt-attr-macro", default-features = false }
+usdt-impl = { path = "../usdt-impl", default-features = false, version = "0.3.4" }
+usdt-macro = { path = "../usdt-macro", default-features = false, version = "0.3.4" }
```
gjcolombo, line 15: One more thought: Should the version dependencies here be exact (i.e. `= 0.3.4` instead of just `0.3.4`) so that floating to a new patch version is not allowed? (This was what bit me with the Propolis change--pinning back to usdt 0.3.2 wasn't sufficient to keep it from grabbing usdt-macro 0.3.3 and getting the incorrect usdt-impl dependency.)

usdt/Cargo.toml
bnaecker: Will do, good idea.

usdt/Cargo.toml
bnaecker: Ah, nice catch, thank you. Incoming!

usdt/Cargo.toml
bnaecker: Done in 1d10020

usdt/Cargo.toml
bnaecker: Done in 1d10020

gjcolombo:
New changes LGTM!

Issue #66, opened 2022-11-07
Compilation fails on Ubuntu Jammy aarch64


alfa07:
On master the compilation fails with:
```
error: invalid register `al`: unknown register
   --> usdt-impl/src/record.rs:116:17
    |
116 |                 inout("al") version,
    |                 ^^^^^^^^^^^^^^^^^^^

error: could not compile `usdt-impl` due to previous error

```

```
ubuntu@u-22-04-jammy-main:~/usdt$ uname -a
Linux u-22-04-jammy-main 5.15.0-52-generic #58-Ubuntu SMP Thu Oct 13 08:04:57 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```
```
ubuntu@u-22-04-jammy-main:~/usdt$ git log
commit eac0fe5f03c3fbf23468ead5cb140f62d51ac3f3 (HEAD -> master, origin/master, origin/HEAD)
Author: Patrick Mooney <pmooney@pfmooney.com>
Date:   Wed Apr 20 20:07:46 2022 -0500

    Handle ASM features more gracefully

```
```
ubuntu@u-22-04-jammy-main:~/usdt$ rustc --version
rustc 1.65.0 (897e37553 2022-11-02)
```

bnaecker:
Thanks for reporting this @alfa07. Looks like you're building on an ARM, which is only really supported on macOS at the moment. I have a fix from earlier in the year, that I seem to have forgotten to finalize and push, which extends support by naming the right registers on other ARM platforms. I'll take a look at that soon, and try to get this sorted.

So I understand better, are you building this crate directly, or as part of a dependency of something else? We've only done minimal testing on Linux platforms, given that DTrace is not very well-supported. (If at all, I'm not sure of the state.) To be clear, you should definitely be able to build this crate and inspect the resulting binaries, but I just want to make sure you are aware of the limitations on Linux!

alfa07:
@bnaecker thank you for looking into it, yes, I am building on macOS 12.6 (Chip: Apple M1 Max) laptop inside ubuntu container :)

bnaecker:
Closed by #79. @alfa07 Let me know if doesn't resolve your problems. I don't have an ARM system available for testing, but the offending lines of code are no longer arch-specific so it _should_ be resolved. Thanks!

Issue #59, opened 2022-04-20
Record parsing could be lazier


bnaecker:
While parsing a probe record, we currently [read the entire object file into memory](https://github.com/oxidecomputer/usdt/blob/c32ec7af3394eee0fbc5c539724f8a89e8a6be48/usdt-impl/src/record.rs#L51). This could be lazier, memory-mapping the file instead and allowing the VM system to lazily load the pages we need. I've used the [`memmap`](https://docs.rs/memmap/latest/memmap/) crate with success previously, though there's also [this crate](https://crates.io/crates/mmap-rs). A manual wrapper may also be fine, since the use case is well-constrained.

bnaecker:
Closed by #79

Issue #58, opened 2022-04-20
Atomic CAS of probe record versions is spicy


bnaecker:
To prevent loading and processing a probe record multiple times, we do a [compare-and-swap](https://github.com/oxidecomputer/usdt/blob/c32ec7af3394eee0fbc5c539724f8a89e8a6be48/usdt-impl/src/record.rs#L209) to exchange a sentinel version number of 255 with the actual version. This is a bit...spicy...since the data being modified is referred to through a shared reference, a `&[u8]` of the probe record bytes. We should probably handle this a bit differently, using a unique mutable reference to the underlying data.

pfmooney:
If probe registration was synchronized with a module wide mutex, then we could probably do away with the atomics.  In order to preserve the idempotency, we could still change the version fields in the probe section, just through a `ptr` write, rather than a reference.

bnaecker:
@ahl We originally decided on the CAS operations via assembly, but I don't remember much of the detail around that decision. Do you have opinions about either of these approaches, a module-wide mutex or raw pointers? The former is probably easier, meaning we'd only have to replaces the CAS functions with an update of that static variable. The latter means more extensive changes to the `record` module, replacing the current `&[u8]` with something like a `*mut u8` and length, but would avoid locking.

pfmooney:
I think if you're updating the version, regardless of whether its via atomic or not, that the parsing logic needs to use pointers rather than references.  Changing data covered by an immutable reference is UB, unless I'm mistaken.

bnaecker:
Agreed. In the case of the former, I assumed we wouldn't update the version numbers in the records themselves. The indication that the records have already been processed would be in something like an `AtomicBool`.

ahl:
I don't remember this discussion in detail, but I do recall being concerned that there might be multiple, distinct uses of the usdt crate within a single binary. I realize now that this might reflect a confused understanding of how cargo works...

bnaecker:
Closed by #79

PR #79, opened 2022-12-19
Reworks probe record version manipulation

bnaecker:
- The `dusty` binary now `mmap`s the target file, rather than reading the whole thing in. This trades a bit of speed in the form of page faults, for laziness, and only reads the required portions of the file.
- Reworks how we read and potentially update the probe record version stored in the target file. First, adds an argument to the `process_section` function controlling whether we want to bother updating the version. `dusty` does not set this, whereas the code that registers the probes does. Second, makes the bytes in which the probe records are stored mutable everywhere, which properly communicates the intent that these may be updated. Lastly, we avoid further inline assembly and some portability issues by using an atomic rather than a `lock xchg` instruction. That currently works by transmuting the byte storing the probe record version, but may work with no unsafe code when the `AtomicU8::from_mut` method is stabilized. In either case, the code should be sound at this point, since writing to that data is now done through an exclusive reference.

### DIFF
--- a/dusty/src/main.rs
+++ b/dusty/src/main.rs
@@ -14,11 +14,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::fs;
-use std::path::{Path, PathBuf};
-
 use dof::Section;
 use goblin::Object;
+use memmap::Mmap;
+use memmap::MmapOptions;
+use std::fs::File;
+use std::fs::OpenOptions;
+use std::path::Path;
+use std::path::PathBuf;
 use structopt::StructOpt;
 use usdt_impl::Error as UsdtError;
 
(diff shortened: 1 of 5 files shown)


bnaecker:
This should also fix #66, by removing the `asm!` invocation that referred to an x86 register. By using `AtomicU8`, there's no dependency on the target arch at all.

bnaecker:
Friendly ping on this @ahl

usdt-impl/src/no-linker.rs
```diff
@@ -138,9 +138,9 @@ fn extract_probe_records_from_section() -> Result<Section, crate::Error> {
     let data = unsafe {
         let start = (&dtrace_probes_start as *const usize) as usize;
         let stop = (&dtrace_probes_stop as *const usize) as usize;
-        std::slice::from_raw_parts(start as *const u8, stop - start)
+        std::slice::from_raw_parts_mut(start as *mut u8, stop - start)
     };
-    process_section(data)
+    process_section(data, /* register = */ true)
```
ahl, line 143: FWIW rust-analyzer drops in the `register` annotation

dusty/src/main.rs
```diff
@@ -66,15 +69,23 @@ fn main() {
 }
 
 // Extract probe records from the given file, if possible.
-pub(crate) fn probe_records<P: AsRef<Path>>(file: P) -> Result<Section, UsdtError> {
-    let data = fs::read(file)?;
-    let section = locate_probe_section(&data).ok_or(UsdtError::InvalidFile)?;
+pub(crate) fn probe_records<P: AsRef<Path>>(path: P) -> Result<Section, UsdtError> {
+    let file = OpenOptions::new()
+        .read(true)
+        .write(true)
```
ahl, line 75: this is scary (ish?) we're opening the on-disk file RW so that we can treat it as `mut`... but we pinky-swear that we're not going to modify the file! because `register` is `false`! for realsies!

But if we're going to use `unsafe` anyways, why not just map it RO and lie about it being `mut`?

ahl (commented):
looks good! nits.

usdt-impl/src/no-linker.rs
bnaecker: You mean that RA adds that comment for you? That's helpful! I don't use RA, so thought it'd be useful for humans as well. Is it more harmful than helpful?

dusty/src/main.rs
bnaecker: Talked through this offline. We can use a private, COW mapping, and then open the file RO. That should never write through to disk, even if we hold this all incorrectly.

bnaecker:
This should resolve the following:

- #58 We're now passing a mutable array of bytes all the way down to the point of an atomic CAS, so this no longer applies
- #59 We mmap the file now, which has the benefit of being slightly lazier. It turns out we usually need to read the bulk of the file to find the probe records anyway, but in some circumstances it will result in less IO. Note that this comes at the expense of speed, since we trade reading the whole file up front for the faults we take when needing to read in the next chunk of the file.
- #64 is obsoleted by this. Many thanks to the author of that PR for bringing this up and proposing solutions. We've adopted some of them here.
- #66 We've removed the inline assembly the previously implemented the probe version atomic CAS. That means we don't rely on the target arch, and so should compile on `aarch64` just fine.

PR #132, opened 2023-05-08
Support usize and isize as native integer types

### LINKED ISSUE #131: `usize` probe arguments are turned into `char *`
I had a couple probes with `usize` as a probe argument:
```
fn serial_uart_write(n: usize) {}
```

When consuming those probes with a D script, I noticed that the type of the probe argument was a `char *` instead of a 64-bit unsigned value:
```
dtrace: failed to compile script ./scripts/serial-console.d: line 59: printf( ) argument #2 is incompatible with conversion #1 prototype:
        conversion: %lu
         prototype: unsigned long
          argument: char *
```

Looking at the source a bit, I think this is because `usize` isn't handled as a [simple type](https://github.com/oxidecomputer/usdt/blob/ac02c2ceb3d592c1df0bac5435aa03d5e8b8f1f4/usdt-attr-macro/src/lib.rs#L295-L303) when the probe arguments are parsed. In any event, it doesn't seem right that `usize` is turned into `char *`.

bnaecker:
Resolves #131


bnaecker:
I added a few tests for this. We can `cargo expand` the new test in `./tests/usize/src/main.rs`, and see that the emitted assembly doesn't serialize the arguments as JSON anymore. We can also see that the probe definition maps a `usize` and `isize` to `uint64_t` and `int64_t` respectively on my 64-bit machine:

```
bnaecker@feldspar : ~/usdt $ dusty ./target/debug/usize
Section {
    ident: Ident {
        magic: [
            127,
            68,
            79,
            70,
        ],
        model: LP64,
        encoding: LittleEndian,
        version: 2,
        dif_vers: 2,
        dif_ireg: 8,
        dif_treg: 8,
    },
    providers: {
        "usize-test": Provider {
            name: "usize-test",
            probes: {
                "emit_isize": Probe {
                    name: "emit_isize",
                    function: "_ZN3syn4expr7parsing14trailer_helper17h046c1dc3b2d10d31E",
                    address: 7631432,
                    offsets: [
                        142,
                    ],
                    enabled_offsets: [
                        0,
                    ],
                    arguments: [
                        "int64_t",
                    ],
                },
                "emit_u8": Probe {
                    name: "emit_u8",
                    function: "_ZN3syn4expr7parsing14trailer_helper17h046c1dc3b2d10d31E",
                    address: 7631508,
                    offsets: [
                        141,
                    ],
                    enabled_offsets: [
                        0,
                    ],
                    arguments: [
                        "uint8_t",
                    ],
                },
                "emit_usize": Probe {
                    name: "emit_usize",
                    function: "_ZN3syn4expr7parsing14trailer_helper17h046c1dc3b2d10d31E",
                    address: 7631418,
                    offsets: [
                        80,
                    ],
                    enabled_offsets: [
                        0,
                    ],
                    arguments: [
                        "uint64_t",
                    ],
                },
            },
        },
    },
}
```

And the real test is that we can actually print the arguments as integers, not strings:

```
bnaecker@feldspar : ~/usdt $ pfexec dtrace -Zn 'usize*::: { printf("%d", arg0); }' -c ./target/debug/usize
dtrace: description 'usize*::: ' matched 0 probes
dtrace: pid 8439 has exited
CPU     ID                    FUNCTION:NAME
 10   5097 _ZN5usize4main17ha57d683e90393598E:emit_usize 1
 10   5095 _ZN5usize4main17ha57d683e90393598E:emit_isize 1
 10   5096 _ZN5usize4main17ha57d683e90393598E:emit_u8 1
```

That `%d` format string would fail to compile if these were still `char*`s.

ahl (approved):
looks good; a lot more code than I anticipated!

jordanhendricks (approved):
Thanks for addressing this! I appreciate that this is a more correct approach than what I gestured at (namely, using the system pointer size).

PR #135, opened 2023-08-10
add tokio + DTrace docs

ahl:
@bnaecker per our discussion

### DIFF
--- a/docs/tracing-tokio.md
+++ b/docs/tracing-tokio.md
@@ -1,0 +1,75 @@
+# Tracing `tokio`
+
+Many DTrace examples you'll find make extensive use of thread-local variables, prefixed with
+`self->`. For example, you'll often see scripts like this:
+
+```dtrace
+// !!! Don't do this with tokio!!!
+
+pid$target::my_func:entry
+{
+	self->arg = arg0;
+}
+
+pid$target::my_func:return
+/self->arg/
+{
+	// Do something with self->arg
+	self->arg = 0;
+}
+```
+
+This uses the system concept of a thread, and--while `async` and `tokio` docs go to great lengths to
+liken a "task" to a "thread"--these are fundamentally different entities! In particular, an `async`
+Rust executor (such as `tokio`) will often multiplex tasks over a thread pool. This means that the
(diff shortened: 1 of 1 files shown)


docs/tracing-tokio.md
```diff
@@ -0,0 +1,75 @@
+# Tracing `tokio`
+
+Many DTrace examples you'll find make extensive use of thread-local variables, prefixed with
+`self->`. For example, you'll often see scripts like this:
+
+```dtrace
+// !!! Don't do this with tokio!!!
+
+pid$target::my_func:entry
```
bnaecker, line 9: Maybe name this `my_async_func` to drive the point home?

docs/tracing-tokio.md
```diff
@@ -0,0 +1,75 @@
...
+
+This uses the system concept of a thread, and--while `async` and `tokio` docs go to great lengths to
+liken a "task" to a "thread"--these are fundamentally different entities! In particular, an `async`
+Rust executor (such as `tokio`) will often multiplex tasks over a thread pool. This means that the
+system's notion of a thread (as in the example above) is meaningless with regard to tracing `async`.
+
+We can think of DTrace's thread-local variables as shorthand for a global array indexed by the
+current thread. So `self->foo` is (conceptually) equivalent to `foo[curthread]`. Since a task may be
+run on various threads, the current thread isn't a useful key.
+**In order to correlate activies across calls, you need some other unique ID**.
```
bnaecker, line 30: "activity across _async_ calls"?

docs/tracing-tokio.md
```diff
@@ -0,0 +1,75 @@
...
+{
+	// Trace stuff of interest
+	event[arg0] = 0;
+}
+```
+
+Note that this D uses a **global** associative array whose key is the unique ID. We do **not** use
+thread-local variables because (as noted exhaustively) the task may run on other threads and other
+tasks may run on this thread.
+
```
bnaecker, line 74: Looks excellent. Might be worth one quick note that this only applies to async runtimes like `tokio`, which do in fact move tasks between threads? Not all of them do, I think, though `tokio` is by far the most common.

bnaecker (commented):
Super super helpful. Thanks!

PR #138, opened 2023-12-10
Add JSON support to dusty

zephraph:
Having just recently learned out useful dtrace is, I'm on a bit of a mission to make what we can do with it more visible. The first part in that story is making it easier to programmatically munge the output of dusty, thus a new JSON mode.

I updated `dof` to include serde serialization of the `Section` struct and its children. I did make a breaking change to `fmt_dof` to have it take in an enum struct called `FormatMode` instead of the flags it was previously taking. The flags were somewhat misrepresentative because `include_sections` was only relevant when in raw mode. Overall I just felt this a better mechanism of invocation, especially when considering adding a new flag.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -105,6 +105,7 @@ jobs:
           --verbose
           --workspace
           --exclude compile-errors
+          --exclude dusty
 
   recent-nightly-no-support:
     name: Run tests on a recent nightly
@@ -129,6 +130,7 @@ jobs:
           --exclude does-it-work
           --exclude test-json
           --exclude test-unique-id
+          --exclude dusty
 
   old-nightly:
     name: Run tests on an older nightly
@@ -150,6 +152,7 @@ jobs:
           --verbose
           --workspace
           --exclude compile-errors
+          --exclude dusty 
 
   old-nightly-no-support:
     name: Run tests on an older nightly
@@ -174,3 +177,4 @@ jobs:
           --exclude does-it-work
           --exclude test-json
           --exclude test-unique-id
+          --exclude dusty
--- a/dof/Cargo.toml
+++ b/dof/Cargo.toml
@@ -1,8 +1,10 @@
 [package]
 name = "dof"
 version = "0.1.5"
-authors = ["Benjamin Naecker <ben@oxidecomputer.com>",
-           "Adam H. Leventhal <ahl@oxidecomputer.com>"]
+authors = [
+  "Benjamin Naecker <ben@oxidecomputer.com>",
+  "Adam H. Leventhal <ahl@oxidecomputer.com>",
+]
 edition = "2018"
 license = "Apache-2.0"
 description = "Tools to read and write the DTrace Object Format (DOF)"
@@ -13,6 +15,8 @@ goblin = { version = "0.6", optional = true, features = ["elf64", "mach64"] }
 pretty-hex = { version = "0.3", optional = true }
 thiserror = "1"
 zerocopy = "0.6"
+serde = "1"
+serde_json = "1"
 
 [features]
 des = ["pretty-hex", "goblin"]
--- a/dof/src/dof.rs
+++ b/dof/src/dof.rs
@@ -25,6 +25,7 @@ use std::{
     convert::{TryFrom, TryInto},
 };
 
+use serde::Serialize;
 use thiserror::Error;
 
 // Magic bytes for a DOF section
@@ -56,7 +57,7 @@ pub enum Error {
 }
 
 /// Represents the DTrace data model, e.g. the pointer width of the platform
-#[derive(Debug, Clone, Copy, serde::Serialize)]
+#[derive(Debug, Clone, Copy, Serialize)]
 #[repr(u8)]
 pub enum DataModel {
     None = 0,
@@ -87,7 +88,7 @@ impl TryFrom<u8> for DataModel {
 }
 
 /// Represents the endianness of the platform
-#[derive(Debug, Clone, Copy, serde::Serialize)]
+#[derive(Debug, Clone, Copy, Serialize)]
 #[repr(u8)]
 pub enum DataEncoding {
     None = 0,
@@ -118,7 +119,7 @@ impl TryFrom<u8> for DataEncoding {
 }
 
 /// Static identifying information about a DOF section (such as version numbers)
-#[derive(Debug, Clone, Copy, serde::Serialize)]
+#[derive(Debug, Clone, Copy, Serialize)]
 pub struct Ident {
     pub magic: [u8; 4],
     pub model: DataModel,
@@ -174,7 +175,7 @@ impl Ident {
 }
 
 /// Representation of a DOF section of an object file
-#[derive(Debug, Clone, serde::Serialize)]
+#[derive(Debug, Clone, Serialize)]
 pub struct Section {
     /// The identifying bytes of this section
     pub ident: Ident,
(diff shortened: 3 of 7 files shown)


dusty/src/main.rs
```diff
@@ -39,15 +39,28 @@ struct Cmd {
     /// Print raw binary data along with summaries or headers
     #[structopt(short, long)]
     raw: bool,
+
+    /// Format output as JSON. Not compatible with `--raw`.
+    #[structopt(long)]
+    json: bool,
```
zephraph, line 45: Technically these are mutually exclusive flags. I looked through structopt's docs and I didn't see a way of indicating mutual exclusivity. I could convert this to clap and do it that way. Technically that'd be the _right_ way, especially since `structopt` is more or less deprecated for clap 3.

zephraph: Went ahead and did the clap upgrade

.github/workflows/rust.yml
```diff
@@ -105,6 +105,7 @@ jobs:
           --verbose
           --workspace
           --exclude compile-errors
+          --exclude dusty
```
zephraph, line 108: Clap requires `1.70.0` essentially which means dusty won't work in these older nightly version tests. I figure since it's mostly auxiliary to the other tooling here then that's likely fine. If that isn't the case, let me know and I can revert the switch to clap and other related changes.

dusty/Cargo.toml
```diff
@@ -5,10 +5,11 @@ authors = ["Benjamin Naecker <ben@oxidecomputer.com>"]
 edition = "2018"
 description = "Tool to inspect USDT probe records in object files"
 license = "Apache-2.0"
+rust-version = "1.70.0"
```
zephraph, line 8: Just explicitly added this given that there are multiple clap deps that require 1.70.0 as a minimum.

.github/workflows/rust.yml
bnaecker: We can bump these toolchains. They're pinned so that we have consistent output for the tests that verify error messages and other compiler outputs, but bumping them and updating the tests is perfectly fine.

dof/Cargo.toml
```diff
@@ -1,8 +1,10 @@
 [package]
 name = "dof"
 version = "0.1.5"
-authors = ["Benjamin Naecker <ben@oxidecomputer.com>",
-           "Adam H. Leventhal <ahl@oxidecomputer.com>"]
+authors = [
+  "Benjamin Naecker <ben@oxidecomputer.com>",
+  "Adam H. Leventhal <ahl@oxidecomputer.com>",
+]
```
ahl, line 7: delete

dof/src/dof.rs
```diff
@@ -56,7 +56,7 @@ pub enum Error {
 }
 
 /// Represents the DTrace data model, e.g. the pointer width of the platform
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, serde::Serialize)]
```
ahl, line 59: any reason you're not doing `use serde::Serialize` ?

dof/src/dof.rs
```diff
@@ -193,6 +193,11 @@ impl Section {
     pub fn as_bytes(&self) -> Vec<u8> {
         crate::ser::serialize_section(self)
     }
+
+    /// Serialize a section into a JSON representation of the DOF object file section.
+    pub fn to_json(&self) -> String {
+        serde_json::to_string(self).unwrap()
```
ahl, line 199: to_string_pretty?

dusty/src/main.rs
```diff
@@ -22,32 +23,43 @@ use std::fs::File;
...
     file: PathBuf,
 
     /// Operate more verbosely, printing all available information
-    #[structopt(short, long)]
+    #[arg(short, long)]
     verbose: bool,
 
     /// Print raw binary data along with summaries or headers
-    #[structopt(short, long)]
+    #[arg(short, long, conflicts_with = "json")]
```
bnaecker, line 39: I think a `--format` argument that accepts the `FormatMode` directly might be more flexible? One would need to add a `FromStr` implementation for `FormatMode`, but that should be it.

dof/src/fmt.rs
```diff
@@ -92,39 +92,62 @@ fn fmt_dof_sec_type<T: Debug + FromBytes + Copy>(data: &[u8]) -> String {
...
-/// headers. If false, only the section headers are printed.
+/// Uses the `FormatMode` to determine how the data is formatted.
 ///
 /// If the file is not of the correct format, or has invalid DOF data, an `Err` is returned. If the
 /// file has no DOF data, `None` is returned.
 pub fn fmt_dof<P: AsRef<Path>>(
     file: P,
-    raw: bool,
-    include_sections: bool,
+    format: FormatMode,
```
zephraph, line 117: As stated in the description, I could've just added another bool flag here but these arguments have semantic relation and switching to an enum seemed the more practical long term solution. In the previous case `raw: false, include_sections: true` had no impact on the output which was no explicit from function's API. 

This _is_ a breaking change for the `dof` crate, but I feel it's a worth while one.

bnaecker (commented):
Thanks, this looks good! I actually have a slightly bigger ask. Could we move `dusty` into the `usdt` package itself? E.g., `git mv ./dusty/src/main.rs ./usdt/src/bin/dusty.rs`. We could then make its main function `probe_records()` in `dusty/src/main.rs` part of the public library interface of `usdt`, which would make other programmatic interactions easier without needing to run a subprocess.

ahl (approved):
looks good to me

dof/src/fmt.rs
bnaecker: That all seems fine to me, I have no issues changing this interface. I do think we should bump the version of the crate though, when we publish. I think it's on `0.1.x`, so we should go to `0.2.0`, probably?

dof/Cargo.toml
zephraph: Delete the authors or delete the change? I think it's only changed due to autoformatting.

dof/src/dof.rs
zephraph: Nope, I can do that.

dof/src/dof.rs
zephraph: Can do. I had that initially but I figured folks could always use jq, etc

zephraph:
> Thanks, this looks good! I actually have a slightly bigger ask. Could we move dusty into the usdt package itself? E.g., git mv ./dusty/src/main.rs ./usdt/src/bin/dusty.rs. We could then make its main function probe_records() in dusty/src/main.rs part of the public library interface of usdt, which would make other programmatic interactions easier without needing to run a subprocess.

@bnaecker I don't think we should move dusty into usdt, but that doesn't mean the `probe_records` fn couldn't itself be moved and then just reimported in dusty. My strong argument _against_ moving dusty into usdt is one of dependencies. Dusty relies on clap (now) which itself _isn't_ necessarily a small dependency. The crate isolation boundary here feels like it makes sense and I'm hesitant to blur those lines. 

I'd also ask that we defer moving functionality from dusty into usdt to another PR. I can see where that added behavior would be useful in `usdt` but it seems outside the scope of this change.

dusty/src/main.rs
zephraph: Perhaps, yeah, I just feel like it's maybe not as terse. IF we did that, we'd want to wrap `FormatMode` in a struct and define that in this module. If you're okay with it I'd just like to move forward with this shape and we can re-evaluate if/when another state needs to be added.

.github/workflows/rust.yml
zephraph: Could we defer this to another PR?

bnaecker:
@zephraph Sure, I can do those other parts in another PR. I get the desire for slim dependencies, and `clap` ain't that. I also think we can work around that with features. The main reason for doing that is easy `cargo install`-ability, but that might not be as valuable as I'm imagining.

PR #139, opened 2023-12-12
Move probe_records to usdt

zephraph:
As suggested by @bnaecker in #138, this pulls `probe_records` out of dusty and moves it into `usdt` to make it easier to programmatically extract the probes from a binary without spawning dusty as a subprocess. 

I _didn't_ add this as a feature but I could see how one might want to add it as a feature given it pulls in several more dependencies.


bnaecker (approved):
Looks good, thanks!

PR #140, opened 2023-12-13
Bump all editions, and required versions for new release

bnaecker:


bnaecker:
It's an incompatible version bump because the `dof` API changed slightly. The `usdt` crate version is also incompatible since the `dof` crate is a transitive dep.

dof/Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [package]
 name = "dof"
-version = "0.1.5"
+version = "0.2.0"
 authors = [
   "Benjamin Naecker <ben@oxidecomputer.com>",
   "Adam H. Leventhal <ahl@oxidecomputer.com>",
 ]
```
ahl, line 7: I'm all for removing these

ahl (commented):
looks good to adam

PR #143, opened 2023-12-15
Update `usdt::probe_records` to work on macOS

zephraph:
After failing to get an xtask experiment @bnaecker was working on I realized that when we extracted out `probe_records` from dusty we missed a key piece of functionality that enabled it to work on osx.

As I was working on this I realized we really needed to rethink how formatting itself was happening. Now there's only one formatting logic branch _and_ probe_records should work upstream in the xtask.

The one thing to note is that `probe_records` now returns a `Vec<Section>` where it previously only returned a `Section`.

### DIFF
--- a/usdt/src/lib.rs
+++ b/usdt/src/lib.rs
@@ -348,7 +348,7 @@
 //! [serde]: https://serde.rs
 //! [asm-sym-feature-pr]: https://github.com/rust-lang/rust/pull/90348
 
-use dof::Section;
+use dof::{extract_dof_sections, Section};
 use goblin::Object;
 use memmap::{Mmap, MmapOptions};
 use std::fs::{File, OpenOptions};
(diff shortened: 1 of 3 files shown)


usdt/src/lib.rs
```diff
@@ -446,14 +446,22 @@ pub fn register_probes() -> Result<(), Error> {
...
+
+    // File contains no DOF data. Try to parse out the ASM records inserted by the `usdt` crate.
     let file = OpenOptions::new().read(true).create(false).open(path)?;
     let (offset, len) = locate_probe_section(&file).ok_or(Error::InvalidFile)?;
 
     // Remap only the probe section itself as mutable, using a private
     // copy-on-write mapping to avoid writing to disk in any circumstance.
     let mut map = unsafe { MmapOptions::new().offset(offset).len(len).map_copy(&file)? };
-    usdt_impl::record::process_section(&mut map, /* register = */ false)
+    usdt_impl::record::process_section(&mut map, /* register = */ false).map(|s| vec![s])
```
ahl, line 464: It's interesting that we don't do both: look for .SUNW_dof sections and for our rust special sections... (nothing you need to change for this PR)

ahl (approved):
Are we just testing this by hand on macos and illumos?

bnaecker (approved):

PR #153, opened 2024-01-02
Use cargo-release and bump minor versions

bnaecker:


bnaecker:
I'm getting set up with `cargo release` to help address #8. We don't really need a tool for managing all this yet, since this commit is literally just `cargo release version minor --execute`. We may want to exclude the test crates from this process, but I don't see much harm in bumping those at the same time.

usdt/Cargo.toml
```diff
@@ -14,14 +14,14 @@ repository = "https://github.com/oxidecomputer/usdt.git"
...
+usdt-impl = { path = "../usdt-impl", default-features = false, version = "=0.5.0", features = [
   "des",
 ] }
-usdt-macro = { path = "../usdt-macro", default-features = false, version = "= 0.4.1" }
-usdt-attr-macro = { path = "../usdt-attr-macro", default-features = false, version = "= 0.4.1" }
-dof = { path = "../dof", features = ["des"], version = "= 0.2.0" }
+usdt-macro = { path = "../usdt-macro", default-features = false, version = "=0.5.0" }
+usdt-attr-macro = { path = "../usdt-attr-macro", default-features = false, version = "=0.5.0" }
+dof = { path = "../dof", features = ["des"], version = "=0.3.0" }
 goblin = { version = "0.6", features = ["elf32", "elf64"] }
```
ahl, line 25: noticed that this is pretty down-rev -- do we want to consider updating it at the same time?

ahl (approved):
I'd suggest we set all crates we don't intend to publish to "0.0.0" and put a file like this in the directory: https://github.com/oxidecomputer/progenitor/blob/main/example-build/release.toml

usdt/Cargo.toml
bnaecker: That's part of #152. The bump changed the API enough that I need to manually update things, so I'll take care of that in a separate PR.

bnaecker:
Good idea, thanks @ahl. After ignoring all the internal crates, here's what `cargo release` will do:

```
bnaecker@shale : ~/usdt $ cargo release version minor
   Upgrading dof from 0.3.0 to 0.4.0
    Updating usdt's dependency from =0.3.0 to =0.4.0
    Updating usdt-impl's dependency from =0.3.0 to =0.4.0
    Updating usdt-impl's dependency from =0.3.0 to =0.4.0
   Upgrading dtrace-parser from 0.2.0 to 0.3.0
    Updating usdt's dependency from =0.2.0 to =0.3.0
    Updating usdt-attr-macro's dependency from =0.2.0 to =0.3.0
    Updating usdt-impl's dependency from =0.2.0 to =0.3.0
    Updating usdt-macro's dependency from =0.2.0 to =0.3.0
   Upgrading usdt-impl from 0.5.0 to 0.6.0
    Updating usdt's dependency from =0.5.0 to =0.6.0
    Updating usdt-attr-macro's dependency from =0.5.0 to =0.6.0
    Updating usdt-macro's dependency from =0.5.0 to =0.6.0
   Upgrading usdt-attr-macro from 0.5.0 to 0.6.0
    Updating usdt's dependency from =0.5.0 to =0.6.0
   Upgrading usdt-macro from 0.5.0 to 0.6.0
    Updating usdt's dependency from =0.5.0 to =0.6.0
   Upgrading usdt from 0.5.0 to 0.6.0
warning: aborting release due to dry run; re-run with `--execute`
```

That's what I really want, and still no need for a script or tool to run that.

PR #159, opened 2024-01-08
Fix race in JSON tests

### LINKED ISSUE #155: macOS tests are flaky
See [here](https://github.com/oxidecomputer/usdt/actions/runs/7391202948/job/20107546079) for an example test run. These have become much more flaky recently, so this warrants further investigation. I don't recall these ever failing in CI, but have now run into this a half-dozen times in recent weeks. We may want to do the dumb thing and bump the duration we wait for the DTrace subproc, or there may be a smarter approach.

bnaecker:
The tests for JSON output work by starting DTrace as a subprocess; firing some probes in the Rust test code; and then grabbing the DTrace output and verifying its properties. There was a race here: although we waited to fire the probes until requested, we never actually waited for DTrace to correctly start and indicate readiness. Test hangs ensue, at least on macOS where `sudo` actually requires a password in most cases.

This adds synchronization in the other direction, waiting for DTrace to print a sentinel in its BEGIN clause, so that the tests don't start until it has successfully spawned.

Fixes #155

### DIFF
--- a/tests/test-json/src/main.rs
+++ b/tests/test-json/src/main.rs
@@ -135,8 +135,7 @@ mod tests {
         let now = Instant::now();
         wait_for_begin_sentinel(&mut dtrace, &now).await;
 
-        // We should now see the probe having fired exactly once. Grab the output data, kill the
-        // DTrace process, and then parse the output as JSON.
+        // Instruct the task firing probes to continue.
         tx.send(()).await.unwrap();
 
         // Wait for the process to finish, up to a pretty generous limit.
--- a/tests/test-json/src/main.rs
+++ b/tests/test-json/src/main.rs
@@ -1,6 +1,6 @@
 //! Integration test verifying JSON output, including when serialization fails.
 
-// Copyright 2022 Oxide Computer Company
+// Copyright 2024 Oxide Computer Company
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


tests/test-json/src/main.rs
```diff
@@ -67,102 +67,142 @@ mod tests {
...
+            .expect("Failed to spawn DTrace subprocess");
+
+        // Wait for DTrace to correctly start up before notifying the test thread to start
+        // firing probes.
+        let now = Instant::now();
+        wait_for_begin_sentinel(&mut dtrace, &now).await;
+
+        // We should now see the probe having fired exactly once. Grab the output data, kill the
+        // DTrace process, and then parse the output as JSON.
+        tx.send(()).await.unwrap();
```
ahl, line 140: I don't feel like this comment goes with this code... although I think this is carried over from the previous version.

tests/test-json/src/main.rs
```diff
@@ -67,102 +67,142 @@ mod tests {
...
-    // This is required to make sure DTrace is "ready" to receive the probe, which takes a bit of time
-    // after the process itself starts.
-    const SLEEP_DURATION: Duration = Duration::from_secs(1);
-
     // Maximum duration to wait for DTrace, controlling total test duration
     const MAX_WAIT: Duration = Duration::from_secs(30);
 
-    fn run_test(recv: Receiver<()>) {
+    // A sentinel printed by DTrace, so we know when it starts up successfully.
+    const BEGIN_SENTINEL: &str = "BEGIN";
```
ahl, line 84: missed joke opportunity

ahl (approved):
looks good

tests/test-json/src/main.rs
bnaecker: Ah, yeah I meant to remove that, thanks for catching.

PR #160, opened 2024-01-08
Some dependency updates that require manual work

bnaecker:
- Update to `rstest v0.18.2`, which requires another toolchain bump for a transitive dep. We're now at 1.66 for the older nightly tests.
- Update `zerocopy v0.7.32`, which changed public traits


bnaecker:
This obviates #149 and #156

ahl (approved):

PR #214, opened 2024-04-29
update ISA specific tests for aarch64

### LINKED ISSUE #213: Latest macOS runs appear broken
We've started to see CI failures on all macOS runs in the last week or so. [Here](https://github.com/oxidecomputer/usdt/actions/runs/8881584455/job/24384292247?pr=209) are the logs from an example. It looks like GH's macOS runners were updated recently, from 12.x to 14.x, which may have caused the failures. I cannot repro this on my machine, but that's quite old at this point (mid-2014, x86, macOS 11.7). @ahl can repro this running on an ARM-based machine, with macOS 14.x.

The failures appear to be related to changes in the register names and / or the ASM we emit or parse out from `clang`'s output.

ahl:
Closes #213

### DIFF
--- a/usdt-impl/src/common.rs
+++ b/usdt-impl/src/common.rs
@@ -376,7 +376,10 @@ mod tests {
             })),
             DataType::Native(dtrace_parser::DataType::String),
         ];
-        let registers = &["rdi", "rsi"];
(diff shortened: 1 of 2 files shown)


bnaecker (approved):
Awesome, thanks

PR #252, opened 2024-07-09
Move off archived action-rs/toolchain

bnaecker:


bnaecker:
Some of the GH actions we were using were quite old, generating deprecation warnings. This switches us to modern alternatives, and cleans up the warnings.

ahl (approved):
oh jeez. that's ancient!

Issue #57, opened 2022-04-14
Remove public feature flags


bnaecker:
The current implementation of `usdt` (v0.3.2) uses the public feature flag `"asm"` to decide what code to generate inside the probe macros. We initially did this for a few reasons. First, the Rust `asm` feature was required, and it's pretty common practice to gate nightly features in a crate with a public feature flag. Secondly, it was very useful for development to pick a real or no-op implementation of the probes.

The time has come to remove this feature flag, however. The `asm` feature has been stabilized. The `asm_sym` feature, required for macOS, is on the path to stabilization. More importantly, a feature flag for this doesn't make a whole lot of sense. First, the "real" version of the probes is the whole point of the crate, so turning them off seems silly. Second, feature flags are additive, which I didn't quite appreciate when I built it this way initially. This is a big problem, since it means it's possible for trouble when crate uses two other crates, both of which use `usdt`. If one enables the probes, it doesn't matter whether the other crate chose to enable them, the probes are there due to feature unification.

This issue tracks removing all public feature flags. The crate will choose the code to be generated in the probe macros solely on the basis of the target OS.

bnaecker:
It may be a good time to tackle this. The last feature we really need, `asm_sym`, stabilized some time ago. If we're willing to push the MSRV to 1.66, then we can use a stable toolchain on all supported platforms and avoid any nightly messiness or feature checks. That MSRV is not too far away from the 1.65 currently implied by some of our dependencies. (E.g., see [this checksuite](https://github.com/oxidecomputer/usdt/actions/runs/7451609092/job/20273078988?pr=156) run.)

It would be very nice to get rid of all that cruft and the attendant build.rs files. 1.66 is pretty old now, so this feels like a reasonable choice.

PR #267, opened 2024-07-29
[meta] update MSRV to 1.75, turn on asm support unconditionally

sunshowers:
There were a lot of build flags working around asm incompatibilities. Now that
inline asm has been stable for a while, we can rip it all out.

Also update the MSRV to 1.75.


usdt-impl/src/no-linker.rs
```diff
@@ -82,10 +82,7 @@ fn compile_probe(
     let (unpacked_args, in_regs) = common::construct_probe_args(&probe.types);
     let is_enabled_rec = emit_probe_record(&provider.name, &probe.name, None);
     let probe_rec = emit_probe_record(&provider.name, &probe.name, Some(&probe.types));
-    #[cfg(usdt_stable_asm)]
-    let asm_macro = quote! { std::arch::asm };
-    #[cfg(not(usdt_stable_asm))]
-    let asm_macro = quote! { asm };
+    let asm_macro = quote! { ::std::arch::asm };
```
pfmooney, line 85: style nit: Could probably just make the direct macro invocation in the `quote!`ed region below, rather than doing an extra assignment?

usdt-impl/src/linker.rs
```diff
@@ -179,10 +179,7 @@ fn compile_probe(
     #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
     compile_error!("USDT only supports x86_64 and AArch64 architectures");
 
-    #[cfg(usdt_stable_asm)]
-    let asm_macro = quote! { std::arch::asm };
-    #[cfg(not(usdt_stable_asm))]
-    let asm_macro = quote! { asm };
+    let asm_macro = quote! { ::std::arch::asm };
```
pfmooney, line 182: style nit: Could probably just make the direct macro invocation in the `quote!`ed region below, rather than doing an extra assignment?

usdt/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 license = "Apache-2.0"
 description = "Dust your Rust with USDT probes"
 repository = "https://github.com/oxidecomputer/usdt.git"
-rust-version = "1.63.0"
+rust-version = "1.75.0"
```
pfmooney, line 9: Commenting here, rather than lines 32-35, because GH review is bad...

We should be able to get rid of the `asm` feature in the child crates (usdt-impl, usdt-macro, usdt-attr-macro), I think?  Leaving it at the top level makes sense, so that consumers are not forced to update their consumption until such time that we cut a version for breaking changes.

sunshowers: Oh good point. I wasn't actually sure whether those features were private or not so I left them in place. Sounds like it's just `usdt` that's public, and the rest are private.

probe-test-build/build.rs
```diff
@@ -17,16 +17,6 @@ use usdt::Builder;
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
 
-    if !version_check::is_min_version("1.59").unwrap_or(false) {
```
bnaecker, line 20: Can we now remove `version_check` entirely, if we don't need it here?

usdt/Cargo.toml
bnaecker: Yeah, all the crates besides `usdt` are private implementation details, so I agree with Patrick, it'd be good to remove them. When we do get around to cutting a v0.6.0 and dropping the public features, we can close #57.

bnaecker (approved):
Thanks for this, it's great cleanup! A couple of thoughts, but otherwise LGTM.

usdt-impl/src/no-linker.rs
sunshowers: Done.

usdt/Cargo.toml
sunshowers: Dropped non-public features, and added a comment.

probe-test-build/build.rs
sunshowers: From `build-dependencies`? Yeah, good point -- done.

sunshowers:
@pfmooney @bnaecker one consequence of this would be that selecting the no-op implementation documented here would no longer be possible:

https://github.com/oxidecomputer/usdt/blob/d4b8090f2bf8527bf6aa0c43e94063ee68647c41/usdt/src/lib.rs#L312-L344

Could you confirm that this is what we want?

sunshowers:
(re-requesting review per q above)

bnaecker (approved):
I think this looks good to me. Being able to select the no-op implementation was mostly a way to let libraries decide whether to trade-off a nightly compiler in exchange for probes. That's no longer needed, so emitting the right probe bodies unconditionally seems like the right call. Those are nearly free so long as the probe is disabled (just a load and compare), so it seems reasonable to always emit them.

Thanks for doing this!

sunshowers:
Thanks -- landed this, happy to address any remaining concerns in followups.

PR #275, opened 2024-08-05
[meta] update pinned toolchain to Rust 1.80

sunshowers:
Also improved CI to ensure that we test on the MSRV and on the latest stable release.

With this we run:

* most tests on MSRV + stable
* trybuild tests on the pinned version

ensuring good coverage across our supported matrix.

Supersedes (and includes) https://github.com/oxidecomputer/usdt/pull/268.

### DIFF
--- a/usdt-impl/build.rs
+++ b/usdt-impl/build.rs
@@ -27,6 +27,9 @@ enum Backend {
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rustc-check-cfg=cfg(usdt_backend_noop)");
+    println!("cargo:rustc-check-cfg=cfg(usdt_backend_linker)");
+    println!("cargo:rustc-check-cfg=cfg(usdt_backend_standard)");
 
     let backend = match env::var("CARGO_CFG_TARGET_OS").ok().as_deref() {
         Some("macos") => Backend::Linker,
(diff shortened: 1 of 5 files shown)


usdt-impl/build.rs
```diff
@@ -27,6 +27,9 @@ enum Backend {
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rustc-check-cfg=cfg(usdt_backend_noop)");
+    println!("cargo:rustc-check-cfg=cfg(usdt_backend_linker)");
+    println!("cargo:rustc-check-cfg=cfg(usdt_backend_standard)");
```
sunshowers, line 32: @pfmooney I didn't gate these here because I updated the pinned `rust-toolchain` as well, ensuring that a simple `cargo build` won't produce any warnings (which is what I was concerned about).

`cargo +1.75.0 build` will produce warnings, but I think that's okay (and it avoids the version-check dependency which we just removed :) )

pfmooney: Fair enough

sunshowers:
@pfmooney @bnaecker if this change looks good to you, could you also update the required CI jobs? There's a new column in the matrix and it looks like the settings need to be updated for that.

(I guess the way to do it would be to land bypassing checks, then to update the required jobs.)

pfmooney (approved):

bnaecker:
Thanks @sunshowers! This process for adding new checks always seems so weird to me. I'll disable the checks temporarily, land this, and then re-enable them all on `main`.

PR #340, opened 2024-11-30
feat: Support Linux stap v3 probes

aapoalas:
This PR makes the crate capable of generating SystemTap's version 3 probes, also sometimes known as SDTs (no relation to DTrace's USDTs, obviously!). This is very nice on Linux systems that don't have any easy way of installing or building DTrace (such as I-use-ArchLinux-by-the-way). This works on all three of the mechanisms for converting D probe definitions into Rust. It also automatically generates the stap versions if `isenabled` probes for each probe.

The open questions in this PR as I see them are:
1. Is there any interest in upstreaming this in the first place?
2. Currently `test_does_it_work` and all tests similar to it do not work, as those rely on DTrace to work. On a generic Linux something similar might be doable with perhaps readelf, or perf, or one of the thousand different eBPF toolsets. What would make the most sense?
3. The methods added for generating GNU assembler argument format are fairly ugly, and I'm not at all sure if they really belong on the DataType enums or if they should go into the `stap3.rs` file.
4. What should I call the usdt-impl? `stap3`? `systemtap`? `ebpf`? `linux`? `knockoff`?
5. Do the `isenabled` probes make sense here? As I understand it, with DTrace USDT's the `isenabled` boolean actually lives in the instruction stream, whereas for SDT's the `isenabled` boolean is read from a mutable static. That sounds like it has a fairly real effect on runtime. If that is the case then it sounds like one would want to give the programmer a way to skip the `isenabled` probe when it isn't needed.

I understand that Oxide doesn't really have skin in the game for expanding the scope this crate to support non-DTrace probes. I'm hoping that you'd still find this potentially worth upstreaming (once the open questions and any issues you have with this PR are resolved): This crate seems to me to be unconditionally the superior way to inject USDTs into Rust code, and it would be great to spread the joy of tracing wider into the Rust ecosystem.

Hoping to hear from you all. Cheers.

### DIFF
--- a/dtrace-parser/src/lib.rs
+++ b/dtrace-parser/src/lib.rs
@@ -87,6 +87,11 @@ pub struct Integer {
 const RUST_TYPE_PREFIX: &str = "::std::os::raw::c_";
 
 impl Integer {
+    pub const POINTER: Self = Self {
+        sign: Sign::Unsigned,
+        width: BitWidth::Pointer,
+    };
+
     fn width_to_c_str(&self) -> &'static str {
         match self.width {
             BitWidth::Bit8 => "8",
@@ -108,6 +113,93 @@ impl Integer {
             _ => "",
         };
         format!("{prefix}int{}_t", self.width_to_c_str())
+    }
+
+    pub fn to_asm_op(&self, i: u8) -> String {
+        // See common.rs for note on argument passing and maximum supported
+        // argument count.
+        #[cfg(target_arch = "x86_64")]
+        if cfg!(target_arch = "x86_64") {
+            let base = match i {
+                0 => "a",
+                1 => "b",
+                2 => "c",
+                3 => "sp",
+                4 => "bp",
+                5 => "di",
+                _ => unreachable!("Up to 6 probe arguments are currently supported"),
+            };
+            let prefix = match self.width {
+                BitWidth::Bit8 => "",
+                BitWidth::Bit16 => "",
+                BitWidth::Bit32 => "e",
+                BitWidth::Bit64 => "r",
+                #[cfg(target_pointer_width = "32")]
+                BitWidth::Pointer => "e",
+                #[cfg(target_pointer_width = "64")]
+                BitWidth::Pointer => "r",
+                #[cfg(not(any(target_pointer_width = "32", target_pointer_width = "64")))]
+                BitWidth::Pointer => compile_error!("Unsupported pointer width"),
+            };
+            let suffix = match self.width {
+                BitWidth::Bit8 => "l",
+                BitWidth::Bit16 | BitWidth::Bit32 | BitWidth::Bit64 | BitWidth::Pointer => {
+                    if i < 3 || i == 7 {
+                        // rax, rbx, rcx, [...], rdx
+                        "x"
+                    } else {
+                        // but [rsp, rbp, rdi, rsi]
+                        ""
+                    }
+                }
+            };
+            format!("%{}{}{}", prefix, base, suffix)
+        } else if cfg!(target_arch = "aarch64") {
+            let base = match i {
+                0 => "0",
+                1 => "1",
+                2 => "2",
+                3 => "3",
+                4 => "4",
+                5 => "5",
+                _ => unreachable!("Up to 6 probe arguments are currently supported"),
+            };
+            let prefix = match self.width {
+                BitWidth::Bit8 => "w",
+                BitWidth::Bit16 => "w",
+                BitWidth::Bit32 => "w",
+                BitWidth::Bit64 => "x",
+                #[cfg(target_pointer_width = "32")]
+                BitWidth::Pointer => "w",
+                #[cfg(target_pointer_width = "64")]
+                BitWidth::Pointer => "x",
+                #[cfg(not(any(target_pointer_width = "32", target_pointer_width = "64")))]
+                BitWidth::Pointer => compile_error!("Unsupported pointer width"),
+            };
+            format!("%{}{}", prefix, base)
+        } else {
+            unreachable!()
+        }
+    }
+
+    pub fn to_asm_size(&self) -> i8 {
+        let prefix = match self.sign {
+            Sign::Unsigned => 1,
+            _ => -1,
+        };
+        let size = match self.width {
+            BitWidth::Bit8 => 1,
+            BitWidth::Bit16 => 2,
+            BitWidth::Bit32 => 4,
+            BitWidth::Bit64 => 8,
+            #[cfg(target_pointer_width = "32")]
+            BitWidth::Pointer => 4,
+            #[cfg(target_pointer_width = "64")]
+            BitWidth::Pointer => 8,
+            #[cfg(not(any(target_pointer_width = "32", target_pointer_width = "64")))]
+            BitWidth::Pointer => compile_error!("Unsupported pointer width"),
+        };
+        prefix * size
     }
 
     pub fn to_rust_ffi_type(&self) -> String {
@@ -241,12 +333,31 @@ impl TryFrom<&Pairs<'_, Rule>> for DataType {
 }
 
 impl DataType {
-    /// Convert a type into its C type represenation as a string
+    /// Convert a type into its C type representation as a String
     pub fn to_c_type(&self) -> String {
         match self {
             DataType::Integer(int) => int.to_c_type(),
             DataType::Pointer(int) => format!("{}*", int.to_c_type()),
             DataType::String => String::from("char*"),
+        }
+    }
+
+    /// Convert a type and register index to its GNU assembler operation as a
+    /// String
+    pub fn to_asm_op(&self, i: u8) -> String {
+        match self {
+            DataType::Integer(int) => int.to_asm_op(i),
+            // Integer pointers are dereferenced by wrapping into parentheses.
+            DataType::Pointer(int) => format!("({})", int.to_asm_op(i)),
+            DataType::String => Integer::POINTER.to_asm_op(i),
+        }
+    }
+
+    /// Convert a type to its GNU assembler size representation as an i8
+    pub fn to_asm_size(&self) -> i8 {
+        match self {
+            DataType::Integer(int) => int.to_asm_size(),
+            DataType::Pointer(_) | DataType::String => Integer::POINTER.to_asm_size(),
         }
     }
 
--- a/usdt-impl/src/common.rs
+++ b/usdt-impl/src/common.rs
@@ -117,18 +117,20 @@ fn shared_slice_elem_type(reference: &syn::TypeReference) -> Option<&syn::Type> 
     }
 }
 
+// x86_64 passes the first 6 arguments in registers, with the rest on the stack.
+// We limit this to 6 arguments in all cases for now, as handling those stack
+// arguments would be challenging with the current `asm!` macro implementation.
+#[cfg(target_arch = "x86_64")]
+pub(crate) const ABI_REGS: [&str; 6] = ["rdi", "rsi", "rdx", "rcx", "r8", "r9"];
+#[cfg(target_arch = "aarch64")]
+pub(crate) const ABI_REGS: [&str; 6] = ["x0", "x1", "x2", "x3", "x4", "x5"];
+#[cfg(not(any(target_arch = "aarch64", target_arch = "x86_64")))]
+compile_error!("USDT only supports x86_64 and ARM64 architectures");
+
 // Return code to destructure a probe arguments into identifiers, and to pass those to ASM
 // registers.
 pub fn construct_probe_args(types: &[DataType]) -> (TokenStream, TokenStream) {
-    // x86_64 passes the first 6 arguments in registers, with the rest on the stack.
-    // We limit this to 6 arguments in all cases for now, as handling those stack
-    // arguments would be challenging with the current `asm!` macro implementation.
-    #[cfg(target_arch = "x86_64")]
-    let abi_regs = ["rdi", "rsi", "rdx", "rcx", "r8", "r9"];
-    #[cfg(target_arch = "aarch64")]
-    let abi_regs = ["x0", "x1", "x2", "x3", "x4", "x5"];
-    #[cfg(not(any(target_arch = "aarch64", target_arch = "x86_64")))]
-    compile_error!("USDT only supports x86_64 and ARM64 architectures");
+    let abi_regs = &ABI_REGS;
 
     assert!(
         types.len() <= abi_regs.len(),
@@ -136,7 +138,7 @@ pub fn construct_probe_args(types: &[DataType]) -> (TokenStream, TokenStream) {
     );
     let (unpacked_args, in_regs): (Vec<_>, Vec<_>) = types
         .iter()
-        .zip(&abi_regs)
+        .zip(abi_regs)
         .enumerate()
         .map(|(i, (typ, reg))| {
             let arg = format_ident!("arg_{}", i);
--- a/usdt-impl/src/empty.rs
+++ b/usdt-impl/src/empty.rs
@@ -17,6 +17,7 @@
 // limitations under the License.
 
 use crate::common;
+use crate::record::emit_probe_record;
 use crate::{Probe, Provider};
 use proc_macro2::TokenStream;
 use quote::quote;
--- a/usdt-impl/src/stap3.rs
+++ b/usdt-impl/src/stap3.rs
@@ -1,0 +1,202 @@
+//! The SystemTap probe version 3 of the USDT crate.
+//!
+//! Used on Linux platforms without DTrace.
+
+// Copyright 2021 Oxide Computer Company
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::{common, DataType};
+use crate::{Probe, Provider};
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
+use std::convert::TryFrom;
+
+pub fn compile_provider_source(
+    source: &str,
+    config: &crate::CompileProvidersConfig,
+) -> Result<TokenStream, crate::Error> {
+    let dfile = dtrace_parser::File::try_from(source)?;
+    let providers = dfile
+        .providers()
+        .into_iter()
+        .map(|provider| {
+            let provider = Provider::from(provider);
+            // Ensure that the name of the module in the config is set, either by the caller or
+            // defaulting to the provider name.
+            let config = crate::CompileProvidersConfig {
+                provider: Some(provider.name.clone()),
+                probe_format: config.probe_format.clone(),
+                module: match &config.module {
+                    None => Some(provider.name.clone()),
+                    other => other.clone(),
+                },
+            };
+            compile_provider(&provider, &config)
+        })
+        .collect::<Vec<_>>();
+    Ok(quote! {
+        #(#providers)*
+    })
+}
+
+pub fn compile_provider_from_definition(
+    provider: &Provider,
+    config: &crate::CompileProvidersConfig,
+) -> TokenStream {
+    compile_provider(provider, config)
+}
+
+fn compile_provider(provider: &Provider, config: &crate::CompileProvidersConfig) -> TokenStream {
+    let probe_impls = provider
+        .probes
+        .iter()
+        .map(|probe| compile_probe(provider, probe, config))
+        .collect::<Vec<_>>();
+    let module = config.module_ident();
+    quote! {
+        pub(crate) mod #module {
+            #(#probe_impls)*
+        }
+    }
+}
+
+fn emit_probe_record(prov: &str, probe: &str, types: Option<&[DataType]>) -> String {
+    let section_ident = r#".note.stapsdt, "", "note""#;
+    let arguments = types.map_or_else(String::new, |types| {
+        types
+            .iter()
+            .enumerate()
+            .map(|(reg_index, typ)| {
+                // Argument format is Nf@OP, N is -?{1,2,4,8} for sign and bit
+                // width, f is for floats, @ is a separator, and OP is the
+                // "actual assembly operand".
+                format!("{}@{}", typ.to_asm_size(), typ.to_asm_op(reg_index as u8))
+            })
+            .collect::<Vec<_>>()
+            .join(" ")
+    });
+    let sema_name = format!("__usdt_sema_{}_{}", prov, probe);
+    format!(
+        r#"
+        // First define the semaphore
+            .ifndef {sema_name}
+                    .pushsection .probes, "aw", "progbits"
+                    .weak {sema_name}
+                    .hidden {sema_name}
+            {sema_name}:
+                    .zero 2
+                    .type {sema_name}, @object
+                    .size {sema_name}, 2
+                    .popsection
+            .endif
+        // Second define the is_enabled probe which uses the semaphore
+                    .pushsection {section_ident}
+                    .balign 4
+                    .4byte 992f-991f, 994f-993f, 3    // length, type
+            991:
+                    .asciz "stapsdt"        // vendor string
+            992:
+                    .balign 4
+            993:
+                    .8byte 990b             // probe PC address
+                    .8byte _.stapsdt.base   // link-time sh_addr of base .stapsdt.base section
+                    .8byte {sema_name}      // link-time address of the semaphore variable, zero if no associated semaphore
+                    .asciz "{prov}"         // provider name
+                    .asciz "{probe}"        // probe name
+                    .asciz ""               // is_enabled probe takes no parameters
+            994:
+                    .balign 4
+                    .popsection
+        // Third define the actual DTrace probe
+                    .pushsection {section_ident}
+                    .balign 4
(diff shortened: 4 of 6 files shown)


ahl:
I'll respond in more detail but: really excited that you've taken this up and we are definitely interested in adding this functionality. As you say, it's not a top priority for Oxide but we'll carve out some time as we can to check this out

dtrace-parser/src/lib.rs
```diff
@@ -110,6 +115,93 @@ impl Integer {
         format!("{prefix}int{}_t", self.width_to_c_str())
     }
 
+    pub fn to_asm_op(&self, i: u8) -> String {
```
ahl, line 118: is this the right place for this? I think the point of this crate is to parse DTrace files..

usdt-impl/src/stap3.rs
```diff
@@ -0,0 +1,184 @@
+//! The SystemTap probe version 3 of the USDT crate.
+//!
+//! Used on Linux platforms without DTrace.
```
ahl, line 3: I think we usually put this below the copyright and license stuff?

ahl:
> 1. Is there any interest in upstreaming this in the first place?

Yes. We would love it if users of this crate could produce useful probes on Linux. This is--in part--why we eschewed the name `dtrace` and chose `usdt` instead.

> 2. Currently `test_does_it_work` and all tests similar to it do not work, as those rely on DTrace to work. On a generic Linux something similar might be doable with perhaps readelf, or perf, or one of the thousand different eBPF toolsets. What would make the most sense?

I think something in the eBPF universe would make sense? Something that a typical user might employ. Do you have suggestions?

> 3. The methods added for generating GNU assembler argument format are fairly ugly, and I'm not at all sure if they really belong on the DataType enums or if they should go into the `stap3.rs` file.

My inclination would be to keep as much of the specific code in the stap3.rs file (and other implementation-specific files such as no-linker.rs) as appropriate.

> 4. What should I call the usdt-impl? `stap3`? `systemtap`? `ebpf`? `linux`?

I think stap3 makes as much sense as anything. Can you walk through consideratwions?

> 5. Do the `isenabled` probes make sense here? As I understand it, with DTrace USDT's the `isenabled` boolean actually lives in the instruction stream, whereas for SDT's the `isenabled` boolean is read from a mutable static. That sounds like it has a fairly real effect on runtime. If that is the case then it sounds like one would want to give the programmer a way to skip the `isenabled` probe when it isn't needed.

I see: so with the stap implementation always performs a load from that mutable static? This seems.. nuts? Can I possibly be understanding this properly??

> 6. Is it intended that a Rust probe provider function defined as `&u32` gets translated into a probe taking a `u32` by value? See [this test](https://github.com/oxidecomputer/usdt/blob/master/usdt-attr-macro/src/lib.rs#L464).

Yes... I think. @bnaecker ?

> I understand that Oxide doesn't really have skin in the game for expanding the scope this crate to support non-DTrace probes. I'm hoping that you'd still find this potentially worth upstreaming (once the open questions and any issues you have with this PR are resolved): This crate seems to me to be unconditionally the superior way to inject USDTs into Rust code, and it would be great to spread the joy of tracing wider into the Rust ecosystem.

Are there other crates people use for Linux / eBPF / Stap / etc?

ahl (commented):
good progress. is there a good overview of the mechanism? Would it be appropriate to include an overview of how this thing works or links out? I can't recall what we have in linker.rs and no-linker.rs but certainly that would be warranted there as well.

bnaecker:
> > Is it intended that a Rust probe provider function defined as &u32 gets translated into a probe taking a u32 by value? See [this test](https://github.com/oxidecomputer/usdt/blob/master/usdt-attr-macro/src/lib.rs#L464).
>
> Yes... I think. @bnaecker ?

Yes, this is mostly the case. We do consider that probe to ultimately have a 32-bit, unsigned integer type, not a pointer. When we emit the probe macro itself, we try to borrow the input, but we then dereference the borrow anyway, passing a copy. For these native integer types, it doesn't matter so much since everything fits in a register. We take more care with strings to pass a pointer, but we also need to null-terminate the data anyway, which necessitates another copy.

Hope that answers your questions, but happy to dig more into the details if needed!

dtrace-parser/src/lib.rs
bnaecker: I agree, this should probably be moved to the `usdt-impl` crate. All the types we need to operate on from the `dtrace-parser` crate are public, so we have access to everything we need there.

usdt-impl/src/empty.rs
```diff
@@ -17,6 +17,7 @@
 // limitations under the License.
 
 use crate::common;
+use crate::record::emit_probe_record;
```
bnaecker, line 20: This looks unused, let's remove if so.

usdt-impl/src/common.rs
```diff
@@ -117,26 +117,28 @@ fn shared_slice_elem_type(reference: &syn::TypeReference) -> Option<&syn::Type>
...
-    // x86_64 passes the first 6 arguments in registers, with the rest on the stack.
-    // We limit this to 6 arguments in all cases for now, as handling those stack
-    // arguments would be challenging with the current `asm!` macro implementation.
-    #[cfg(target_arch = "x86_64")]
-    let abi_regs = ["rdi", "rsi", "rdx", "rcx", "r8", "r9"];
-    #[cfg(target_arch = "aarch64")]
-    let abi_regs = ["x0", "x1", "x2", "x3", "x4", "x5"];
-    #[cfg(not(any(target_arch = "aarch64", target_arch = "x86_64")))]
-    compile_error!("USDT only supports x86_64 and ARM64 architectures");
+    let abi_regs = &ABI_REGS;
```
bnaecker, line 133: I think we can just inline this into the `.zip()` call, as `&ABI_REGS`.

dtrace-parser/src/lib.rs
bnaecker: Could we also provide a bit more documentation as to the purpose of this function? What does it actually produce and how is it used?

dtrace-parser/src/lib.rs
```diff
@@ -110,6 +115,93 @@ impl Integer {
...
+                BitWidth::Bit32 => "e",
+                BitWidth::Bit64 => "r",
+                #[cfg(target_pointer_width = "32")]
+                BitWidth::Pointer => "e",
+                #[cfg(target_pointer_width = "64")]
+                BitWidth::Pointer => "r",
+                #[cfg(not(any(target_pointer_width = "32", target_pointer_width = "64")))]
+                BitWidth::Pointer => compile_error!("Unsupported pointer width"),
+            };
+            let suffix = match self.width {
```
bnaecker, line 144: I think it would be more clear to merge the matches for computing both `prefix` and `suffix`. Right now, the conditional logic on L147 is a bit obtuse, and I think the duplication is worth being very explicit about what prefix / suffix is used when.

usdt-impl/src/stap3.rs
```diff
@@ -0,0 +1,184 @@
...
+
+// Copyright 2021 Oxide Computer Company
```
bnaecker, line 5: I think this is probably copy-paste. I'm actually not sure _who_ owns the copyright here (probably you, but IANAL), but the year should be current in any case.

usdt-impl/src/stap3.rs
```diff
@@ -0,0 +1,184 @@
...
+    let section_ident = r#".note.stapsdt, "", "note""#;
+    let arguments = types.map_or_else(String::new, |types| {
+        types
+            .iter()
+            .enumerate()
+            .map(|(reg_index, typ)| {
+                // Argument format is Nf@OP, N is -?{1,2,4,8} for sign and bit
+                // width, f is for floats, @ is a separator, and OP is the
+                // "actual assembly operand".
+                format!("{}@{}", typ.to_asm_size(), typ.to_asm_op(reg_index as u8))
```
bnaecker, line 85: Is there any documentation that describes the ABI for these STap probe records? Either for this argument description format, or the semaphore or actual probe assembler contents?

bnaecker (commented):
Thanks for taking this on, we're certainly excited about adding a new platform!

Most of the code looks pretty reasonable to me. I have a few questions about documentation and other small suggestions, but the big thing is testing. You asked about the "does it work" test -- we should definitely try to replicate that on Linux. That test works by calling out to the `dtrace` command-line utility, and checking that the probes actually fire with the expected data. Is there a similar CLI tool for SystemTap?

aapoalas:
> 2. I think something in the eBPF universe would make sense? Something that a typical user might employ. Do you have suggestions?

I will have to look into this a bit. I'd like to use just readelf but it might be that this will need to be SystemTap. The reason being that `does_it_work` relies on the `register_probes` method working. On Linux what one would usually do is use some CLI command to scan probes out of an executable. Also, as is tradition, at least I am not aware of there being a singular registry of USDT probes and instead each program might well have its own. If that is the case, then my guess would be that SystemTap is the one most likely to have a `/dev/dtrace/helper` equivalent somewhere.

If no equivalent filesystem based API for registering the probes can be found, then the Linux-equivalent test needs to be written in an entirely different way where it depends its own executable path... Though actually now that I think about it, that's probably not too bad? Anyway, I'll look into this.

> 4. I think stap3 makes as much sense as anything. Can you walk through considerations?

So I've been using [this official document](https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation) and [this header-only implementation](https://github.com/libbpf/usdt/blob/46c09f9440257639afc9f8cbf9dcb51f01dd9929/usdt.h#L356-L385) as my sources. `stap3` was my first idea based on the official documentation since it mentions "Only version 3 is described here." And of course "stap" comes from "SystemTap". Thinking on this now, I might prefer changing this to `stapsdt` so that it better matches readelf output (`NT_STAPSDT`): The version number is a minor detail that I doubt is very widely thought about, given that the version 3 [feature request](https://sourceware.org/bugzilla/show_bug.cgi?id=11916) was closed as done in 2010.

`systemtap` would of course be more self-explanatory and have better SEO, for the little that is worth. But, this might also lead people astray into thinking that this won't work with eBPF.

`ebpf` is a "jump on the bandwagon" kind of name. Most Linux tracing seems to prefer eBPF nowadays (when not using plain perf) and this would show the crate to be part of that movement. But, this of course also leads astray by excluding SystemTap and perf.

`linux` would probably be a misleading catch-all. I'd have to dig into the dtrace4linux code to find out what USDT format they use, but I assume there the format is DTrace with minor changes to make the ELF sections work on Linux. If that is the case, then this would obviously be misleading as the correct format for Linux would depend on whether you're using DTrace or perf/SystemTap/eBPF. If dtrace4linux (and Oracle's DTrace on Linux) all use the SystemTap format then this would of course be a perfect option.

**Personally:** I think I'm leaning into `stapsdt`. It's... not self-explanatory, but it doesn't particularly need to be. And it doesn't lead astray in any way.

>5.  I see: so with the stap implementation always performs a load from that mutable static? This seems.. nuts? Can I possibly be understanding this properly??

It does seem a bit nutty, doesn't it? But this is my understanding of the STAPSDT probes. The [documentation](https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation) mentions:

> Semaphores are treated as a counter; your tool should increment the semaphore to enable it, and decrement the semaphore when finished.

A counter in the instruction stream sounds like a bad idea, so this basically has to be in the data. And of course it can't be on the stack so either the heap or statics it is. Between those two I know I'd pick statics any day. The [header-only implementation seems to agree](https://github.com/libbpf/usdt/blob/46c09f9440257639afc9f8cbf9dcb51f01dd9929/usdt.h#L192-L196). [Another Rust impl](https://github.com/cuviper/probe-rs/blob/master/src/platform/systemtap.rs#L70-L76) actually directly uses a `static mut` per call-site, though this [causes problems](https://github.com/bpftrace/bpftrace/issues/3601).

> Are there other crates people use for Linux / eBPF / Stap / etc?

There are at least [sonde](https://github.com/wasmerio/sonde-rs) and [probe](https://github.com/cuviper/probe-rs). Sonde builds a C library that exposes functions that call the actual probes, and calls those functions through FFI from Rust. This is a perfectly working solution and probably gets compiled down to reasonable stuff with LTO or at least PGO, but I'm not a big fan of having to cross the FFI boundary there. Probe generates STAPSDT probes directly the same way usdt does, but the argument handling at least seems to be pretty bare-bones.

usdt-impl/src/stap3.rs
aapoalas: Added documentation links to https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation mostly.

usdt-impl/src/stap3.rs
aapoalas: A lot of this file is copied over from empty and non-linker, so I kept this as Oxide Computer. The argument handling code I moved to a different file where I put the copyright as mine; I don't really give a hoot about that so if you decide you'd prefer it as Oxide Computer then I'll happily change it.

usdt-impl/src/stap3.rs
aapoalas: Fixed.

usdt-impl/src/empty.rs
aapoalas: Leftover code indeed. Fixed.

usdt-impl/src/common.rs
aapoalas: Leftover code from when I originally reused the ABI_REGS in the argument handling. Reverted these changes.

dtrace-parser/src/lib.rs
aapoalas: Merged into a relative huge match statement. Definitely clearer that way: I found that I had bugged the r8 and r9 1 byte, 2 byte and 3 byte names.

dtrace-parser/src/lib.rs
aapoalas: Moved to its own file and added documentation.

aapoalas (commented):
Thank you for the comments! I've addressed them now.

aapoalas:
> You asked about the "does it work" test -- we should definitely try to replicate that on Linux. That test works by calling out to the `dtrace` command-line utility, and checking that the probes actually fire with the expected data. Is there a similar CLI tool for SystemTap?

I was unable to find an easy answer to this: For one, the test is taking advantage of the `register_probes` function writing directly into DTrace's helper file to register the probes. I looked into Linux perf's `perf buildid-cache` command and it has something similar but it's not a single file but has at least a few parameters. I also tried looking into SystemTap itself but was unable to find anything that would directly point towards such a registering API existing.

Of course, even if such a method to register USDTs exists and is shared across all consumers on Linux (SystempTap, perf, eBPF, ...), I'd still need to figure out the data format and do the writing, which I expect would be a fair bit of work as well. Given that eg. [Brendan Gregg's Linux perf USDT notes](https://www.brendangregg.com/perf.html) starts with calling `perf buildid-cache`, I doubt there is really a feasible way to have the program itself register its own probes the way `register_probes` does.

As such, I implemented the `does_it_work` test using `readelf -n current_exe`. It wasn't too bad to do. I'll need to add cfg's to run the test on Linux and skip it otherwise. Right now I just commented out the original test. And then I need to fix all the _other_ tests that do something similar.

aapoalas:
Reminder: I'm waiting for review on this, specifically concerning the review fixes, if going forward with using readelf looks like it make sense to you, and how specializing the tests for Linux should be done.

andreimatei:
Please excuse the drive-by comment from someone trying to understand the USDT ecosystem. 

> If dtrace4linux (and Oracle's DTrace on Linux) all use the SystemTap format then this would of course be a perfect option.

Oracle's DTrace for Linux does not support the SystemTap format (i.e. the `.note.stapsdt` ELF section), does it? Instead, it supports the DTrace Object Format, right?

aapoalas:
> Please excuse the drive-by comment from someone trying to understand the USDT ecosystem.

At least from my point of view, none of that. Thanks for the comment.

> Oracle's DTrace for Linux does not support the SystemTap format (i.e. the `.note.stapsdt` ELF section), does it? Instead, it supports the DTrace Object Format, right?

I'm not sure, but that is probably exactly correct. Linux as usual is then plagued by two competing standards :)

aapoalas:
@ahl Sorry for the long wait; I have finally gotten all tests to pass on Linux! I used bpftrace to observe the probes firing: this required calling bpftrace a little differently compared to DTrace (which shouldn't be a surprise of course; no two APIs can ever be equivalent). Most notably, I had to hook into the running program using the `-p PID` flag in bpftrace.

If you could review this, that'd be swell <3

One consideration regarding the stap probes is whether this should actually be put behind a feature flag, and if so then should Linux default to stapsdt probes, or noop probes?

But in general, I think this should be good; tests pass, I was able to use this to debug a simple test program using bpftrace, and previously I've also used it to debug / trace a much more significant project.

Cheers

ahl (approved):
I mean... seems good? It works?

tests/does-it-work/src/main.rs
```diff
@@ -38,86 +38,218 @@ fn run_test(rx: std::sync::mpsc::Receiver<()>) {
...
-        assert_eq!(
-            second, "args[1]: char *",
-            "Argument is incorrect: {}",
-            second
-        );
-
-        thr.join().expect("Failed to join test runner thread");
+
+    #[cfg(not(target_os = "linux"))]
+    mod unix {
```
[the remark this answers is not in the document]
aapoalas: Good idea <3 Fixed.

usdt-impl/src/common.rs
```diff
@@ -110,7 +110,7 @@ pub fn construct_probe_args(types: &[DataType]) -> (TokenStream, TokenStream) {
     #[cfg(target_arch = "x86_64")]
     let abi_regs = ["rdi", "rsi", "rdx", "rcx", "r8", "r9"];
     #[cfg(target_arch = "aarch64")]
-    let abi_regs = ["x0", "x1", "x2", "x3", "x4", "x5"];
+    let arbi_regs = ["x0", "x1", "x2", "x3", "x4", "x5"];
```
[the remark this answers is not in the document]
aapoalas: Some sort of typo / random button mash  :D Fixed

aapoalas:
> Can we test this is CI as we do for macos and illumos?

I set up the CI to run most tests on Ubuntu as well now. I _hope_ it's correct but luckily CI will tell us that :)

.github/workflows/rust.yml
```diff
@@ -2,9 +2,9 @@ name: Rust
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
```
aapoalas, line 5: **note:** Automatic formatter changes; I can revert these if preferred.

ahl: please do

tests/test-json/src/main.rs
```diff
@@ -176,29 +123,168 @@ mod tests {
                 }
                 _ => {}
             }
-            println!("DTrace not yet ready");
+            // println!("DTrace not yet ready");
```
ahl, line 126: looks like this got commented out; are there other changes?

tests/does-it-work/src/main.rs
```diff
@@ -38,86 +38,218 @@ fn run_test(rx: std::sync::mpsc::Receiver<()>) {
...
+
+    #[cfg(target_os = "linux")]
+    mod stap {
+        use super::run_test;
+        use std::process::Stdio;
+        use std::sync::mpsc::channel;
+        use std::thread;
+
+        #[test]
+        fn test_does_it_work_stapsdt() {
```
ahl, line 137: this is pretty different than the DTrace version; can you explain?

tests/test-json/src/main.rs
```diff
@@ -176,29 +123,168 @@ mod tests {
...
+                ))
+                .arg("-p")
+                .arg(std::process::id().to_string())
+                .stdout(Stdio::piped())
+                .stderr(Stdio::piped())
+                .spawn()
+                .expect("Failed to spawn bpftrace subprocess");
+
+            // Wait for DTrace to correctly start up before notifying the test thread to start
+            // firing probes.
```
ahl, line 247: DTrace?

tests/test-json/src/main.rs
aapoalas: Hmm, this seems like an oversight/mistake; I'll fix this and go over the code once more to make sure I've not left other such unrelated changes in.

tests/test-json/src/main.rs
aapoalas: Nice catch, will fix <3

tests/does-it-work/src/main.rs
aapoalas: Yeah, I originally wrote this using just readelf; I'd like to say I didn't want to dictate a DTrace alternative on Linux CI but honestly I just hadn't used them and thus went with the tool I knew.

I'll rewrite this to use bpftrace to make it match better.

aapoalas: Okay, turns out I might have had more of a thought here than I remembered: bpftrace can output USDTs but will not output their arguments even with the verbose flag. Thus, I left the readelf usage and added a note about this into the test. I also renamed the test to be consistent and simplified its implementation a little to not undeservedly split up statically knowable parts (like checking "Provider:" and "does__it" strings separate instead of just checking for "Provider: does__it").

usdt-impl/src/stapsdt.rs
```diff
@@ -0,0 +1,241 @@
...
+/// let is_enabled = unsafe { __usdt_sema_provider_probe.get().read_volatile() } > 0;
+///
+/// if is_enabled {
+///   let arg1: u64 = get_arg1();
+///   let arg2: u32 = get_arg2();
+///   let arg3: *const u8 = get_arg3();
+///   const {
+///     // Build time: Define this location as being the place in the
+///     // instruction stream where our probe is located.
+///     core::usdt::define::<(u64, u32, *const u8)>("provider", "probe", &__usdt_sema_provider_probe);
```
bnaecker, line 114: The `core::` prefix is a bit misleading. This is a stand-in for the inline assembly below, right? If so, I might remove that.

usdt-impl/src/stapsdt/args.rs
```diff
@@ -0,0 +1,239 @@
...
+/// 1. Read a u64 from RDI: `8@%rdi`.
+/// 2. Read an i32 through a pointer in RSI: `-4@(%rsi)`.
+/// 3. Read an f64 through a pointer in RDI: `8f@(%rdi)`.
+///    (Not sure if `-` should be added.)
+/// 4. Read a u64 through a pointer with an offset: `8%-4(%rdi)`.
+pub(crate) fn format_argument((reg_index, typ): (usize, &DataType)) -> String {
+    format!(
+        "{}@{}",
+        data_type_to_arg_size(typ),
+        data_type_to_asm_op(typ, reg_index as u8)
```
bnaecker, line 237: I'd prefer `u8::try_from(reg_index).unwrap()` here. It's unlikely we'll have > 256 arguments, but I prefer the safety anyway.

bnaecker (commented):
This looks really great, thank you so much for your patience and hard work getting this integrated! I have a few small nits, but overall this looks good. The tests you've added are great.

usdt-impl/src/stapsdt.rs
```diff
@@ -0,0 +1,241 @@
...
+        &probe.name,
+        &provider.use_statements,
+        &probe.types,
+    );
+
+    let sema_name = format_ident!("__usdt_sema_{}_{}", provider.name, probe.name);
+    let impl_block = quote! {
+        {
+            #[repr(C)]
+            struct UsdtSema {
```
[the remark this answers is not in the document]
aapoalas: I'll add a comment about this but basically I'm being overly cautious here: A lot of my learning on how STAPSDTs are done come from this one single-header C library for USDTs and there they [use a C struct for this](https://github.com/libbpf/usdt/blob/46c09f9440257639afc9f8cbf9dcb51f01dd9929/usdt.h#L147). Just in case it could affect anything about anything, I thus decided to also use a `repr(C)` struct. I believe it should be absolutely okay to expose it as a `u16` directly but I was nervous about it since on some platforms a C struct like this will indeed have an effect.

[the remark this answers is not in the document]
aapoalas: Nevermind; I just removed it. Tests (locally) still passed so the overabundance of caution is probably unneeded. (Cue some obscure ELF/compiler magic where structs are actually guaranteed to be 32-bit aligned and STAPSDT semaphores rely on that feature.)

usdt-impl/src/stapsdt/args.rs
```diff
@@ -0,0 +1,239 @@
...
+
+/// Convert an Integer type and a register index into a GNU Assembler operation
+/// that reads the integer's value from the correct register. Effectively this
+/// means generating a string like `%REG` where `REG` is the register that the
+/// data is located in.
+fn integer_to_asm_op(integer: &Integer, reg_index: u8) -> &'static str {
+    // See common.rs for note on argument passing and maximum supported
+    // argument count.
+    if reg_index > 5 {
+        unreachable!("Up to 6 probe arguments are currently supported");
```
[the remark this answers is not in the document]
aapoalas: You're goddamn right! Fixed <3

usdt-impl/src/stapsdt/args.rs
aapoalas: Good point, fixed <3

usdt-impl/src/stapsdt.rs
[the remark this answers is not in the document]
aapoalas: I also went and added an extra comment (and asked in #dark-arts though no responses yet) on it just in case this _does_ lead to something odd and someone needs to come looking for breadcrumbs.

usdt-impl/src/stapsdt.rs
aapoalas: Removed the `core::` prefix and added a bit more comments around here to hopefully make the intent of this pseudo-code a little bit clearer yet.

aapoalas:
> This looks really great, thank you so much for your patience and hard work getting this integrated! I have a few small nits, but overall this looks good. The tests you've added are great.

Addressed all the comments, thank you for them <3 And thank you and ahl for the patience as well <3

PR #341, opened 2024-12-02
Update macOS runner image

bnaecker:


bnaecker:
This uses `macos-15` explicitly to avoid randomly getting macOS 14.7.1, which leads us to #213.

PR #353, opened 2024-12-17
Remove `Clone` requirement for argument types

### LINKED ISSUE #136: Remove unnecessary `Clone` requirement
The docs mention a `Clone` requirement on arbitrary serializable types you'd like to include in a probe, and specifically mentions that the data involved is never actually cloned.

I wonder if the issue is similar to the one the `sqlx` project had for their `PoolOptions` type (see: https://github.com/launchbadge/sqlx/pull/2553), where a `Clone` derive on a type which involves `T` in a generic manner results in an excessive `Clone` requirement on it even though the data is never actually cloned. In this case, the solution was to manually implement `Clone` on the containing type, as shown in the linked pull request.

It's possible this _isn't_ the cause (I haven't yet dug into the code to see), but reading the docs triggered my feeling that it _might_ be a similar issue. I'm happy to tinker (with some guidance on the right place to look) to see if I can remove the `Clone` requirement!

bnaecker:
- Rework construction of the internal type-checking function to remove the clone requirement. Construct and call the type-check-function right at the site we already expand the probe argument closure, rather than outside. This means we don't need to expand things twice, and don't need to clone the closure, and so don't need the argument types to be cloneable either.
- Fixup tests for new method, add some comments.
- Update trybuild tests
- Closes #136

### DIFF
--- a/usdt-impl/src/no-linker.rs
+++ b/usdt-impl/src/no-linker.rs
@@ -82,6 +82,12 @@ fn compile_probe(
     let (unpacked_args, in_regs) = common::construct_probe_args(&probe.types);
     let is_enabled_rec = emit_probe_record(&provider.name, &probe.name, None);
     let probe_rec = emit_probe_record(&provider.name, &probe.name, Some(&probe.types));
+    let type_check_fn = common::construct_type_check(
+        &provider.name,
+        &probe.name,
+        &provider.use_statements,
+        &probe.types,
+    );
 
     let impl_block = quote! {
         {
@@ -91,24 +97,25 @@ fn compile_probe(
                     "990:   clr rax",
                     #is_enabled_rec,
                     out("rax") is_enabled,
-                    options(nomem, nostack, preserves_flags)
+                    options(nostack)
                 );
             }
 
             if is_enabled != 0 {
                 #unpacked_args
+                #type_check_fn
                 unsafe {
                     ::std::arch::asm!(
                         "990:   nop",
                         #probe_rec,
                         #in_regs
-                        options(nomem, nostack, preserves_flags)
+                        options(nostack)
                     );
                 }
             }
         }
     };
-    common::build_probe_macro(config, provider, &probe.name, &probe.types, impl_block)
+    common::build_probe_macro(config, &probe.name, &probe.types, impl_block)
 }
 
 fn extract_probe_records_from_section() -> Result<Section, crate::Error> {
--- a/usdt-impl/src/no-linker.rs
+++ b/usdt-impl/src/no-linker.rs
@@ -97,7 +97,7 @@ fn compile_probe(
                     "990:   clr rax",
                     #is_enabled_rec,
                     out("rax") is_enabled,
-                    options(nostack)
+                    options(nomem, nostack)
                 );
             }
 
@@ -109,7 +109,7 @@ fn compile_probe(
                         "990:   nop",
                         #probe_rec,
                         #in_regs
-                        options(nostack)
+                        options(nomem, nostack, preserves_flags)
                     );
                 }
             }
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -46,7 +46,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ "macos-15" ]
+        os: [ "macos-15", "ubuntu-latest" ]
         # Test on the pinned rust-toolchain version.
         toolchain: [ "1.80.0" ]
     steps:
@@ -78,10 +78,10 @@ jobs:
           --release
           --verbose
           --workspace
+          --exclude compile-errors
           --exclude does-it-work
           --exclude test-json
           --exclude test-unique-id
-          --exclude compile-errors
 
   stable-test-no-op:
     name: Test with probes disabled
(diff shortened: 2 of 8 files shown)


usdt-impl/src/no-linker.rs
```diff
@@ -91,24 +97,25 @@ fn compile_probe(
...
             if is_enabled != 0 {
                 #unpacked_args
+                #type_check_fn
                 unsafe {
                     ::std::arch::asm!(
                         "990:   nop",
                         #probe_rec,
                         #in_regs
-                        options(nomem, nostack, preserves_flags)
+                        options(nostack)
```
ahl, line 112: my read of `nomem` and `preserves_flags` is that they're appropriate here; is there a reason you removed them?

ahl (approved):
looks good; you might try building crucible or something with this just to make sure nothing broke.

usdt-impl/src/no-linker.rs
bnaecker: I was being cautious. The flags might be preserved, I think, but I'm not sure about memory accesses. The raw assembly we've written doesn't read or write any. But let's suppose there's a `char *` in one of the registers, because we're passing a string as a probe argument. DTrace will need to access that memory, but I don't really know if that counts here.

ahl: > I was being cautious. The flags might be preserved, I think, [...]

Yeah. It's really the nop or the mov so flags are preserved.

> but I'm not sure about memory accesses. The raw assembly we've written doesn't read or write any. But let's suppose there's a `char *` in one of the registers, because we're passing a string as a probe argument. DTrace will need to access that memory, but I don't really know if that counts here.

I don't think that's going to count. Say a debugger attaches to this process when the %rip is on this specific instruction; it then does loads and stores. That's fine.

> The asm! block does not read from or write to any memory accessible outside of the asm! block. This allows the compiler to cache the values of modified global variables in registers across the asm! block since it knows that they are not read or written to by the asm!. The compiler also assumes that this asm! block does not perform any kind of synchronization with other threads, e.g. via fences.

Yeah, we're not going to screw up cached loads for example...

bnaecker: Ok, I'll put these flags back.

bnaecker: If I'm understanding correctly, then the `nomem` option is also fine in the first ASM block too, right? That's the `clr %rax` bit, which does potentially touch the flags, but does not access any memory at all.

ahl: That's fair enough; `clr` can impact flags.

bnaecker: I put the `nomem` option back in the first (is-enabled) probe, and kept the original options in the second. Done in [68f4ec7](https://github.com/oxidecomputer/usdt/pull/353/commits/68f4ec761246403ab05c041684ef8ab209a0f812), let me know if you have any other questions @ahl.

bnaecker:
I pulled this branch into a binary with some probes in the Omicron repo, and it looks like the probes still fire as normal:

```
^Cbnaecker@flint : ~/file-cabinet/oxide/omicron/oximeter/db $ sudo dtrace -n 'clickhouse_io*:::' -p 1112 | c++filt
dtrace: description 'clickhouse_io*:::' matched 23 probes
CPU     ID                    FUNCTION:NAME
  3   4366 _ZN156_$LT$oximeter_db..native..io..packet..client..Encoder$u20$as$u20$tokio_util..codec..encoder..Encoder$LT$oximeter_db..nati:packet-sent
  3   4365 _ZN11oximeter_db6native10connection10Connection11query_inner28_$u7b$$u7b$closure$u7d$$u7d$17hef12990ca5dec621E:packet-sent
  0   4363 _ZN104_$LT$oximeter_db..native..io..packet..server..Decoder$u20$as$u20$tokio_util..codec..decoder..Decoder$GT$6decode17h485c565:packet-received
  0   4350 _ZN11oximeter_db6native10connection10Connection11query_inner28_$u7b$$u7b$closure$u7d$$u7d$17hef12990ca5dec621E:data-packet-received
  0   4363 _ZN104_$LT$oximeter_db..native..io..packet..server..Decoder$u20$as$u20$tokio_util..codec..decoder..Decoder$GT$6decode17h485c565:packet-received
  0   4350 _ZN11oximeter_db6native10connection10Connection11query_inner28_$u7b$$u7b$closure$u7d$$u7d$17hef12990ca5dec621E:data-packet-received
  0   4363 _ZN104_$LT$oximeter_db..native..io..packet..server..Decoder$u20$as$u20$tokio_util..codec..decoder..Decoder$GT$6decode17h485c565:packet-received
  0   4363 _ZN104_$LT$oximeter_db..native..io..packet..server..Decoder$u20$as$u20$tokio_util..codec..decoder..Decoder$GT$6decode17h485c565:packet-received
  3   4363 _ZN104_$LT$oximeter_db..native..io..packet..server..Decoder$u20$as$u20$tokio_util..codec..decoder..Decoder$GT$6decode17h485c565:packet-received
  3   4363 _ZN104_$LT$oximeter_db..native..io..packet..server..Decoder$u20$as$u20$tokio_util..codec..decoder..Decoder$GT$6decode17h485c565:packet-received
  3   4350 _ZN11oximeter_db6native10connection10Connection11query_inner28_$u7b$$u7b$closure$u7d$$u7d$17hef12990ca5dec621E:data-packet-received
  3   4363 _ZN104_$LT$oximeter_db..native..io..packet..server..Decoder$u20$as$u20$tokio_util..codec..decoder..Decoder$GT$6decode17h485c565:packet-received
  2   4363 _ZN104_$LT$oximeter_db..native..io..packet..server..Decoder$u20$as$u20$tokio_util..codec..decoder..Decoder$GT$6decode17h485c565:packet-received
```

That's while running some SQL queries against ClickHouse using a little shell I wrote.

ahl:
ship it

PR #451, opened 2025-09-02
fix: remove invalid cfg requirement for x86_64 in integer_to_asm_op

aapoalas:
A mistake I managed to slip in as part of #340: an errant `cfg` (leftovers from a previous version of the code) made the function remove all of its important parts if the machine was not x86_64.

This wasn't caught by CI because none of the CI builds actually run on Arm currently. I added running on macos-latest which are aarch64 machines.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -65,7 +65,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ "windows-latest" ]
+        os: [ "ubuntu-latest", "windows-latest" ]
         # Test on MSRV and stable.
         toolchain: [ "1.85.0", "stable" ]
     steps:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ "macos-15", "ubuntu-latest" ]
+        os: [ "macos-15", "macos-latest", "ubuntu-latest" ]
         # Test on MSRV and stable.
         toolchain: [ "1.85.0", "stable" ]
     steps:
@@ -46,7 +46,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ "macos-15", "ubuntu-latest" ]
+        os: [ "macos-15", "macos-latest", "ubuntu-latest" ]
         # Test on the pinned rust-toolchain version.
         toolchain: [ "1.85.0" ]
     steps:
@@ -88,7 +88,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ "macos-15" ]
+        os: [ "macos-15", "macos-latest" ]
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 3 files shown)


bnaecker:
I'm not sure I understand. `macos-15` is explicitly listed as an Arm-based runner, from [here](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories). Do you mean there's no Arm runner specifically for the SystemTap backend? If that's what we need, we can use the `ubuntu-24.04-arm` runner instead. See [here](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-22-image.md) for a description of that runner.

aapoalas:
> I'm not sure I understand. `macos-15` is explicitly listed as an Arm-based runner, from [here](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories). Do you mean there's no Arm runner specifically for the SystemTap backend? If that's what we need, we can use the `ubuntu-24.04-arm` runner instead. See [here](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-22-image.md) for a description of that runner.

Oof, sorry; don't mind me as I crawl under this rock. Yup; I somehow got mixed up in my [CI job failure](https://github.com/trynova/nova/actions/runs/17376991120/job/49325451484) which is exactly using `ubuntu-24.04-arm` and still picked its next-door neighbour. I'll switch it out.

aapoalas:
Okay; I had to disable the actual "most tests" CI for `ubuntu-24.04-arm`; for some reason though bpftrace is installed, it does not trigger the BEGIN probe in the "does-it-work" test, thus the test hangs and fails.

I don't have an ARM machine myself so I'm left with only the CI to debug that and after staring at it for a while (and checking that bpftrace _is_ actually installed and seeing what the output is while it hangs, a bunch of control characters) I decided to give it up as a bad job.

bnaecker:
It looks like we need to change the branch protection rules, or I can merge even without that particular check. I agree we should remove `ubuntu-latest` from the "no support" test, since it _is_ now supported! I'd also like to add a rule verifying the new FreeBSD support as well. Let me do both of those today.

aapoalas:
> It looks like we need to change the branch protection rules, or I can merge even without that particular check. I agree we should remove `ubuntu-latest` from the "no support" test, since it _is_ now supported! I'd also like to add a rule verifying the new FreeBSD support as well. Let me do both of those today.

Do you want me to revert the revert of the no-support removal? (That's a great phrase right there :D )

aapoalas:
> Yeah, that makes sense.

Reverted; for my shattered pride's sake I beg you remove those lines from the squashed commit message :)

bnaecker:
Oh, I misunderstood. You said the `ubuntu-24.04-arm` tests don't complete, but you don't have a machine to verify locally. I think I'd like to dig into that a bit -- if we're going claim Linux support, and macOS aarch64 support, it'd be nice to at least understand why Linux Arm support is lagging.

aapoalas:
> Oh, I misunderstood. You said the `ubuntu-24.04-arm` tests don't complete, but you don't have a machine to verify locally. I think I'd like to dig into that a bit -- if we're going claim Linux support, and macOS aarch64 support, it'd be nice to at least understand why Linux Arm support is lagging.

Hmm. I'll see if I can't do something with a Raspberry Pi I have lying around...

aapoalas:
> There are definitely differences. The semaphore values are different, as are the arguments. I have no clue if those are _meaningful_ differences.

Ah yeah, those are not meaningful differences, they're just byte offsets into the executable :)

> I think at this point, we should drop our claim Linux Arm support and the branch protection rule requiring the CI tests to pass. I'd like to keep them, as well as all your work doing the actual implementation. We can revisit this another time to see if things have improved, but I don't think it's fair to block this PR any more.
> 
> Thanks for your efforts!

Thank you very much <3 I might circle around to this with a reinstalled Raspberry Pi, but that will have to wait at least a little because... well, grants and things :D

PR #486, opened 2025-10-28
Make linker backend more portable by adding /usr/sbin to PATH.

smolkaj:
This can be helpful because certain build system (like Bazel) do not pass on environment variables in order to make builds more hermetic. Without this change, `usdt` is not compatible with such build systems.

Context: I ran into this issue while adding Bazel BUILD rules to https://github.com/oxidecomputer/p4 (https://github.com/oxidecomputer/p4/pull/196).

### DIFF
--- a/usdt-impl/src/linker.rs
+++ b/usdt-impl/src/linker.rs
@@ -71,6 +71,7 @@ use quote::{format_ident, quote};
 use std::{
     collections::BTreeMap,
     convert::TryFrom,
+    env,
     io::Write,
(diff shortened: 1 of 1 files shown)


ahl:
You're hitting this on macOS I presume?

smolkaj:
Yes exactly.

smolkaj:
Any concerns about this change? Let me know if there is anything I can do to move this forward.

bnaecker (approved):

PR #546, opened 2026-03-26
Fix compiler error on aarch64 no-linker systems

bnaecker:


Mythra:
Hi @bnaecker , I just ran into this issue. If you're looking to install rust, the way I did (and I think the generally recommended way is...):

- Check out ports (if you don't already have them through portsnap/git/etc.): `git clone https://git.freebsd.org/ports.git /usr/ports`
- Go to the rust port: `cd /usr/ports/lang/rust`
- Install the target (will cause a _lot_ of builds/installs to happen): `make install clean`

Hopefully even without CI we can get a fix in for those of using it, I notice it's been open for a bit 😄

bnaecker:
Thanks @Mythra! Does that build all of Rust from scratch? If so, I'm not sure that's the best path forward, running a full build of Rust on every CI run. If that's installing a binary package, I'm definitely more inclined to include it.

Mythra:
Yes it does, if there's no caching solution that yall can implement yourselves.... I would very much not recommend throwing it in the CI hot path.

bnaecker:
Ok, thanks. We can investigate putting a binary package somewhere for these architectures, but that'll have to be a separate PR that doesn't block this one.

